PR #29, opened 2013-07-02
Various Bugs for Windows Listeners - support NTAG and Provide installer

SteveALee:
* Builds with VS 2012
* Single EXE and no extra DLLs (removed libcurl.dll)
* USB listening now works
* No F11 hotkey to avoid clashes

### DIFF
--- a/listeners/GPII_USBListener/README.md
+++ b/listeners/GPII_USBListener/README.md
@@ -1,4 +1,4 @@
-GPII USB User Listener for Windows
+#GPII USB User Listener for Windows
 
 This folder contains the necessary files for the Windows USB listener (source code and libraries).
 
--- a/listeners/README.md
+++ b/listeners/README.md
@@ -1,16 +1,16 @@
-+ Listeners
+# Listeners
 
 The GPII listeners for windows are currently USB and RFID (NFC). See individual README.md files for details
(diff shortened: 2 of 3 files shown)


kaspermarkus:
Hey @SteveALee and @amb26, what's the status of this pull request? Should this stuff still go in, or has it been made obsolete from your listener work in the other repo, steve?

SteveALee:
as far as i'm concerned it should go in. it's the latest listener for USB
devices. Hens spiked a listener to use the internal readers in some devices
like the LGS vaios.

We need to decide on a plan as currently they both use different APIs and
languages

discuss in architecture call today?

Steve Lee
OpenDirective http://opendirective.com

sgithens:
This definately does not replace any work I'm doing. The listener I wrote is specifically for Windows 8 API's and is not supported by a huge amount of RFID hardware yet. (Including our APR122 readers)

sgithens:
@SteveALee I do apologize for the lateness on this.  I will try to fast track.  However, I am unfortunately out next week, and I'm not sure if I'll get to it before I leave on Saturday... If no one else can do I will make it a super high priority the week I get back.  I do have VS and stuff installed so I should be able to test building it.

SteveALee:
Thanks Steve. I'm also away next week for P4A.

Can we also move the Release over to GPII?

Thanks

Steve Lee

PR #59, opened 2015-03-05
Gpii 12

sgithens:
http://issues.gpii.net/browse/GPII-12


amb26:
DDDD! This is an absolutely brilliant PUL request. All it needs to get in is some kind of test case!

What I suggest is - from here http://www.chiark.greenend.org.uk/~sgtatham/putty/download.html - download puttygen.exe (180K), rename it to something unusual, and issue a test case that attempts to kill it, and returns success if it (shortly) disappears from the process list (as reported, for example, by our standard tasklist /fi \"STATUS eq RUNNING\" /FI \"IMAGENAME eq THING command line that is currently cut n pasted into all our acceptance tests....

sgithens:
AAAA! I was wondering a bit about that.  What do you think about a test case for that function just launching and killing an app in windows that is always there such as notepad.exe or maybe a calculator if there is one? Or do you think there is too much worry about closing an app someone has open while developing? (It seems like there must be some innocuous app that always ships with windows we can use to test?)

Also, if I"m correct there are no unit tests yet for WindowsUtilities.js, so I need to add a tests directory for that particular module.

amb26:
Well, I don't think that the behaviour of killing an app that the user may already have running is acceptable  for a test case : P
And I don't think we are entitled to distribute renamed copies of windows builtin apps either. Taking a renamed and very small BSD licenced .exe was the best compromise I could think of.

sgithens:
@amb26 I haven't the tests you recommended.  Please take another look at this review when you have a chance.  Thanks!

SteveALee:
Great work Hens, thanks for this.

sgithens:
@amb26 As requested I've put a comment on that async exec call.

sgithens:
@kaspermarkus Those jshint fixes should be fixed and part of this pull now.  Thanks!

gpii/node_modules/WindowsUtilities/test/testWindowsUtilities.js
```diff
@@ -0,0 +1,57 @@
...
+    var checkForTask = function() {
+        var result = shelljs.exec(tasklistCmd);
+        if (result.output.indexOf(exeName) >= 0) {
+            return true;
+        }
+        else {
+            return false;
+        }
+    };
+    jqUnit.assertDeepEq( "We've haven't started it yet.", false, checkForTask());
```
amb26: This pull should never have been merged. It turns out that all of these assertions are faulty - assertDeepEq will always pass when given a primitive value. This is a QUnit bug. These assertions pass whether they are true or false.

gpii/node_modules/WindowsUtilities/test/testWindowsUtilities.js
```diff
@@ -0,0 +1,57 @@
...
+    //REM On the call below, async is true because if it is false shelljs will
+    //REM wait around until it is manually killed before continuing with the
+    //REM rest of the tests.
+    shelljs.exec(".\\gpii\\node_modules\\WindowsUtilities\\test\\" + exeName, {
+        async: true
+    });
+    jqUnit.assertDeepEq("We just started the new putty.", true, checkForTask());
```
amb26: Given the process starts asynchronously, we never in fact observe it here. But because the assertion is faulty, the test always passes.

gpii/node_modules/WindowsUtilities/test/testWindowsUtilities.js
```diff
@@ -0,0 +1,57 @@
...
+        }
+        else {
+            return false;
+        }
+    };
+    jqUnit.assertDeepEq( "We've haven't started it yet.", false, checkForTask());
+    //REM On the call below, async is true because if it is false shelljs will
+    //REM wait around until it is manually killed before continuing with the
+    //REM rest of the tests.
+    shelljs.exec(".\\gpii\\node_modules\\WindowsUtilities\\test\\" + exeName, {
```
amb26: Because this path is hard-coded, the test fails if it is not run from the top-level driver - for example, it fails if run from its own directory.

gpii/node_modules/WindowsUtilities/test/testWindowsUtilities.js
```diff
@@ -0,0 +1,57 @@
...
+jqUnit.test("Testing Killing Processes", function () {
+    jqUnit.expect(3);
+
+    var exeName = "putty-0FEF1314.exe";
+    var tasklistCmd = "tasklist /fi \"STATUS eq RUNNING\" /FI \"IMAGENAME eq " +
+        exeName + "\"";
+
+    var checkForTask = function() {
+        var result = shelljs.exec(tasklistCmd);
+        if (result.output.indexOf(exeName) >= 0) {
```
amb26: This should simply read 
return result.output.indexOf(exeName) >= 0

gpii/node_modules/WindowsUtilities/test/testWindowsUtilities.js
amb26: JIRAed at https://issues.fluidproject.org/browse/FLUID-5901

SteveALee: nice! :(

PR #61, opened 2015-05-12
NOJIRA: new batch files for creating direct access to PMT and GPII

malarres:


malarres:
related: https://github.com/GPII/universal/pull/360

javihernandez:
Hi @malarres, I've just merged the universal side of this pull request, however, since we don't have a review3 branch of windows, I would say that these scripts should be distributed as a zip file from a gist or something similar. Also, if other pilot sites are willing to use this, it could be documented in our [pilots3 wiki page](http://wiki.gpii.net/w/Cloud4all_Pilot_3_-_Set_up_and_installation). 

@kaspermarkus, do you agree with this?

PR #63, opened 2015-10-04
JIRA-1292: Added new top levell gpii folder to the build instructions

SteveALee:
Otherwise a node-modules is created as peer of windows and that is probably not wanted as will be amongst other project folders


kaspermarkus:
This has been resolved by a change in the windows build script which no longer checks out universal at a higher level folder

PR #72, opened 2016-06-02
GPII-1826: Fix the issues with killProcessByName and spiSettings Handler in node 64bit

javihernandez:
This is a quick fix for terminating processes on Windows 10. See https://issues.gpii.net/browse/GPII-1826. I do know that this isn't the optimal solution (so I don't expect this to land master), but so far it does the work on my windows 10 box.


kaspermarkus:
I chatted with Javi about this yesterday, but for the sake of documentation... It looks like we have licencing issues with pskill, from: https://issues.gpii.net/browse/GPII-12

> By my reading of the Sysinternals licence, http://technet.microsoft.com/en-us/sysinternals/bb469936.aspx - we are not permitted to redistribute the pskill.exe executable. "you may not publish the software for others to copy". I am not a lawyer, but presumably this is clear enough.
> 
> However, using lib-ffi to bridge to this Win32 API:
> 
> http://stackoverflow.com/questions/7956519/win32-api-how-to-kill-processes-by-name
> 
> one would guess would avoid both the permissions as well as licencing issue.

So it looks like we will need a different approach than pskill. Javi and I decided we should probably just discuss it when we're all here in toronto shortly

javihernandez:
In fact I'm not sure if the problem is a 32 vs 64bit thing, something related to privileges or that ffi is not doing the right work. @sgithens, have you experienced this before when working on LGS?

javihernandez:
Ok,

some more info. I've made a quick test and built a simple process killer with the code referenced above and it works. It seems to me that node-ffi is probably not doing it right, but who knows. I'd suggest that we create a simple native addon as part of the killProcess module.

colinbdclark:
@javihernandez I'm so super impressed that you're making so much progress with this! This is awesome.

I'm thinking ultimately we don't want to commit binaries to the repo if we can avoid it. Just so I understand, this executable is based on the work Evgeni did a few years ago under GPII-12? Is the source kicking around somewhere, and is it viable to build it with the same toolchain we use for the rest of the GPII?

@sgithens mentioned today that he has a SUPER SOON TO BE PUSHED Microsoft .NET application that also does this. Hey, @sgithens, can you point us to your stuff so that we can do some kind of comparison between the two and decide what to do?

javihernandez:
> @javihernandez I'm so super impressed that you're making so much progress with this! This is awesome.
> 
> I'm thinking ultimately we don't want to commit binaries to the repo if we can avoid it. Just so I understand, this executable is based on the work Evgeni did a few years ago under GPII-12? Is the source kicking around somewhere, and is it viable to build it with the same toolchain we use for the rest of the GPII?

It uses the very same code from http://stackoverflow.com/questions/7956519/win32-api-how-to-kill-processes-by-name, the only difference is that I'm taking the process' name as a parameter when calling the binary. As I said, I don't expect this to be merged as it is, but I'm sure that we can provide a small native addon to make such magic and build it in the same way we do for linux' modules.

javihernandez:
> @javihernandez Are you running 32-bit or 64-bit Windows 10?
> 
> What I found with the FFI based windows calls to list applications and kill them by name was that it wouldn't actually list processes that were running in 64-bit address space. From looking at various threads on stackoverflow and things, I found similar people problems, but the solution always seemed to be to use a .NET library instead that would simply list all the processes.
> 
> Ideally, I'd like to use this .NET call via edge.js... I apologize that the code on the box I have this on in C# is back home and I won't get to it until Friday... although it was only 20 lines of C# so maybe we could just try and reproduce it this week.

That sounds good, in fact, I'd say that we should think about this, given that we have plans to add support for process reporting in the upcoming months, and we don't have anything for windows (yet) ;)

colinbdclark:
Here's the thing, right? I don't want to spend too long waiting to get our unit tests working again. Do you think this is something we could pair on this week--either implementing a straight up C# utility based on @javihernandez' sketch, or by taking a deeper dive into edge.js?

javihernandez:
> Here's the thing, right? I don't want to spend too long waiting to get our unit tests working again. Do you think this is something we could pair on this week--either implementing a straight up C# utility based on @javihernandez' sketch, or by taking a deeper dive into edge.js?

I guess that we can come up with a solution within this week, but I'm afraid that this is not only thing that break tests at this moment as I'm also having some failures with the spiSettingsHandler. :cry: Probably it's related to this, but I wanted to double check with you while we're in Toronto this week.

sgithens:
@colinbdclark @javihernandez The .NET that takes care of this, so that isn't a problem to knock out quickly... I think it would be very useful for us to spend a little time together taking a look at doing it in edge.js this week as I think it's a very important component for us going forward.

Here are some references to the exceptionally easy to use to .NET calls for this:

http://stackoverflow.com/questions/3345363/kill-some-processes-by-exe-file-name
https://msdn.microsoft.com/en-us/library/system.diagnostics.process.kill(v=vs.110).aspx

javihernandez:
So far, these are the results from testing this problem with different combinations of architectures of Windows and Node.js.

* Windows 10 64bit + Node.js 64bit: **FAIL**
* Windows 10 64bit + Node.js 32bit: **OK**
* Windows 10 32bit + Node.js 32bit: **OK**

I'll give a try to edge.js in order to solve this particular problem, if it doesn't work, I'll create a simple native addon/brigde to make it work.

colinbdclark:
Wow, thanks for the triaging work, @javihernandez! Do all the tests pass (i.e. both the process killing and the SPI tests that were failing) in the environments you've labeled as **OK**?

gloob:
Merged fix in @javihernandez branch to don't lose narrative here https://github.com/javihernandez/windows/pull/1 I would love to see it here :)

javihernandez:
> Merged fix in @javihernandez branch for don't lose narrative here javihernandez#1 I would love to see it here :)

@gloob,

awesome!!! It solves the problem with killing processes in 64bit but still we have issues with the spiSettingsHandler. Have you taken a look at them? How do you feel about getting them fixed? If your expectations are really bad, I'll go ahead and merge https://github.com/javihernandez/windows/pull/2 in order to make nodejs 32bit our default (and "fully supported") arch.

gloob:
As commented in the other thread, merge it meanwhile we finish this, it's going to take a while.

amb26:
We should try to capture some of this discussion on the parent JIRA since I am already a bit lost and we will find it difficult to find this pull in the future. The headline says "use pskill" but it seems that we are no longer doing this?

gloob:
Follow-up in #78

PR #73, opened 2016-06-16
GPII-1826: Switches Node.js to 32-bit version

avtar:
These changes switch the Node.js version back to 32-bit and also clean up the build scripts.

### DIFF
--- a/provisioning/build.bat
+++ b/provisioning/build.bat
@@ -1,3 +1,11 @@
 cd c:\vagrant
-npm install
+call npm install
+
+mkdir c:\node_modules
+cd c:\node_modules
+git clone --depth 1 https://github.com/GPII/universal.git
+cd universal
+call npm install
+call npm install dedupe-infusion
+node -e "require('dedupe-infusion')()"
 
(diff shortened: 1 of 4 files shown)


provisioning/build.bat
```diff
@@ -1,3 +1,11 @@
 cd c:\vagrant
-npm install
+call npm install
+
+mkdir c:\node_modules
+cd c:\node_modules
+git clone --depth 1 https://github.com/GPII/universal.git
+cd universal
+call npm install
+call npm install dedupe-infusion
```
colinbdclark: De-duping Infusion is no longer required. This line and the next can be removed.

sgithens:
Hey!  I'm not terribly familiar with this thread and just saw it pass by on email.  I'm not sure why we're moving back to 32-bit, but in general the node bitness will need to match the Windows bitness if we want the onscreen keyboard to launch

javihernandez:
@sgithens,

what version of windows are you talking about? Just tested this on win10 and worked like a charm.

sgithens:
Just windows 10.  Is the Windows 10 VM that vagrant spins up 32-bit windows 10?

avtar:
@sgithens no, it's Windows 10 64-bit.

PR #74, opened 2016-06-17
GPII-1848: Removal of grunt-gpii and stale build files, update to grunt 1.x

amb26:
and ESLint linting, position universal as a standard npm dependency


Gruntfile.js
```diff
@@ -1,38 +1,31 @@
-/*!
+/*
 GPII Windows Personalization Framework Gruntfile
+Copyright 2013-2014 OCAD University
```
colinbdclark: Why did the copyright statement change here? Was this file imported as-is from Infusion or another project?

Gruntfile.js
```diff
@@ -1,38 +1,31 @@
...
 
-Copyright 2014 RTF-US
+Licensed under the Educational Community License (ECL), Version 2.0 or the New
```
colinbdclark: Why did the license change?

colinbdclark:
A couple of issues with this PR when I tested it using Vagrant:

1. The [provisioning/build.bat](https://github.com/amb26/windows/blob/GPII-1848/provisioning/build.bat) file wasn't updated to reflect this change. We unfortunately had to [switch to this manual script](https://github.com/GPII/windows/commit/806355e3e6dae4d81f322f95de1487f50cd5c5ba) due to issues with Grunt and symlinks on Windows. The grunt-gpii-like workflow in this file should also be removed.
2. When I run try to run the unit or acceptance tests, I get an error saying jqUnit can't be found. If I manually add node-jqunit to the windows repo's <code>devDependencies</code>, I receive an error complaining about multiple infusions and advising me to dedupe.

screen shot 2016-06-17 at 4 49 38 pm

screen shot 2016-06-17 at 4 24 18 pm

Am I missing a step or configuration in order to get the tests to run?

PR #69, opened 2016-04-19
killprocess

dovsons:
kill the child process too

PR #78, opened 2016-07-05
GPII-1826: Fix inconsistencies between different target architectures (32/64 bits)

gloob:
Handle all the problems detected in the unit tests based on problems with the pointers dimension in different architectures.

Basically this patch takes care of the target arch (using os.arch()) and define the pointer based on the arch, also fix one problem with one API method (CreateToolhelp32Snapshot) signature that was creating an error and some of the structures (LOGFONT).

This PR reverts previous commit 32d7eccacfa5509ef90d9543627e82c02de5d361 where we changed node.js to 32 bits.

This PR fix #72 #69 and [GPII-1826](https://issues.gpii.net/browse/GPII-1826)


amb26:
Looks reasonable to me

amb26:
I've closed this pull since I mistakenly merged from your branch named GPII-1826, without noticing that you had generated this pull from the master of your own repository. As far as I can see, the changes are the same, even though the commit hashes don't align, and I was able to make the system pass its tests. Please reopen a pull with any further changes if I have not picked them all up.

gloob:
After rebased with my master everything is good and we can keep this (#78) and #72 closed. Thanks for pushing it @amb26 !

PR #79, opened 2016-07-29
GPII-1873: Fixing problem of Windows defining the value of default scheme to null for HighContrast method.

gloob:
GPII-1873: Fixing problem of Windows defining the value of default scheme to null for HighContrast method.

Workaround for the current problem we are having with HighContrast method.

### DIFF
--- a/gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
+++ b/gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
@@ -249,7 +249,7 @@ gpii.windows.spi.setImpl = function (payload) {
 
     gpii.windows.spi.populateResults(payload, true, false, results);
 
-    GPII1873_HighContrastBug("SET", payload, results);
+    gpii.windows.spi.GPII1873_HighContrastBug("SET", payload, results);
 
     // transform results here   oldValue: x   -->   oldValue: { value: x, path: ... }
     fluid.each(results, function (value, setting) {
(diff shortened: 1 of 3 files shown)


gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
```diff
@@ -296,6 +300,30 @@ fluid.defaults("gpii.windows.spiSettingsHandler.updateCursors", {
...
+ * output of this HIGHCONTRAST methods.
+ *
+ * More information at https://issues.gpii.net/browse/GPII-1873
+ */
+function GPII1873_HighContrastBug(method, payload, results) {
+    if ((payload.options.pvParam.type === "struct") &&
+        (payload.options.pvParam.name === "HIGHCONTRAST")) {
+        if (payload.settings.HighContrastTheme &&
+            payload.settings.HighContrastTheme.value === "") {
+            console.log("Bug #1873 detected. We are forcing the return set to empty string although the system cannot set to it.");
```
amb26: This should be a fluid.log statement . "return set" could better be "return value"

gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
```diff
@@ -296,6 +300,30 @@ fluid.defaults("gpii.windows.spiSettingsHandler.updateCursors", {
...
+ * Fix for GPII-1873.
+ *
+ * This method overrides the output of the system based on the payload information. We
+ * actually know that Windows have a bug that disallow us to reset to the HighContrastTheme
+ * default scheme to null or empty string. That is the reason we are circumventing the
+ * output of this HIGHCONTRAST methods.
+ *
+ * More information at https://issues.gpii.net/browse/GPII-1873
+ */
+function GPII1873_HighContrastBug(method, payload, results) {
```
amb26: Put this function in a public namespace like the other parts of the implementation

gloob:
Thanks @amb26 for reviewing it, all comments addressed in the new commits.

PR #65, opened 2015-10-16
GPII-1000 - Screen resolution setting for windows

SteveALee:
@sgithens @kaspermarkus so now we have a javascript function to call I need to add to the settings handlers. Can you please guide me on best approach? I was sure the test solutions file for win32 used the temp winsetres.exe but can't find it now (that would explain why it now fails in the LGS).

As a general comment It seems the WindowsUtilities exports stuff that could be private or is not used.

Also the unitests include a file higher up in the hierachy "../../../../index.js" that includes other modules. To my mind that is an inversion of dependencies and introduces strong unnecessary coupling. The WindowsUtils tests should only test and depend on the windows utils. Ideally I'd not expect to see any fluid dependencies here but I guess that is a key part of the module system so is necessary.

PR #81, opened 2016-08-02
GPII-1000: Screen resolution setting for windows (bis)

javihernandez:
This includes the original pull request (https://github.com/GPII/windows/pull/65) and is in sync with master.


javihernandez:
Ready for review after adding the universal side of this pr (https://github.com/GPII/universal/pull/465).

amb26:
ok to test

javihernandez:
ok to test

amb26:
OK - this is looking like the more healthy (!) fail that we expected - any ideas what was up with the CI system before, @avtar ?

avtar:
@amb26 Restarting the CI node's instance of VirtualBox was necessary last Friday as it had become unresponsive. Prior to that [a Jenkins bug](https://issues.jenkins-ci.org/browse/JENKINS-36333) was letting anonymous users, including web crawlers, trigger jobs using the ``Resume build`` link. Lots of unnecessary jobs were triggered last Monday and the week prior as a result, causing several runs of unit and acceptance tests on the same VM. @amatas blocked access to that link in the nginx configuration for the CI server to prevent that issue from reoccurring.

javihernandez:
@amb26 this is ready again for another review. And now it passes CI!!
@sgithens this is the new pull request for GPII-1000, which goes with https://github.com/GPII/universal/pull/465. Feel free to leave your comments in here.

gpii/node_modules/displaySettingsHandler/src/displaySettingsHandler.js
```diff
@@ -0,0 +1,211 @@
...
+    ],
+    // https://msdn.microsoft.com/en-us/library/windows/desktop/dd183411(v=vs.85).aspx
+    // DEVMODE*, DWORD
+    "ChangeDisplaySettingsW": [
+        t.LONG, ["pointer", t.DWORD]
+    ]
+});
+
+/**
+    *  Gets the current display's screen resolution
```
sgithens: The formatting on this block comment is over indented.

gpii/node_modules/displaySettingsHandler/src/displaySettingsHandler.js
```diff
@@ -0,0 +1,211 @@
...
+        // note for unknown reason on win 10 the returned dmSize is 188 not expected 220
+        return { width: dm.dmPelsWidth, height: dm.dmPelsHeight };
+    }
+    return { width: 0, height: 0 };
+
+};
+
+
+/**
+    *  Sets the current display's screen resolution if possible
```
sgithens: This formatting on this block comment is also over indented.

gpii/node_modules/displaySettingsHandler/src/displaySettingsHandler.js
```diff
@@ -0,0 +1,211 @@
...
+    [t.DWORD, "dmMediaType"],
+    [t.DWORD, "dmDitherType"],
+    [t.DWORD, "dmReserved1"],
+    [t.DWORD, "dmReserved2"],
+    //#if (WINVER >= 0x0500) || (_WIN32_WINNT >= 0x0400)
+    [t.DWORD, "dmPanningWidth"],
+    [t.DWORD, "dmPanningHeight"]
+    //#endif
+    //#endif
+]);
```
sgithens: Can we get rid of the commented out stuff in this DEVMODEW struct?

javihernandez: yeah, this is something that I spotted and fixed but I stashed the changes at some point and forgot to add them back again. Sorry for that, will add it ASAP.

gpii/node_modules/displaySettingsHandler/src/displaySettingsHandler.js
```diff
@@ -0,0 +1,211 @@
+/*
+ * Windows Display Settings Handler
+ *
+ * Copyright 2012 Raising the Floor - International
```
amb26: I don't think that all of these people were really involved in the creation of this file

gpii/node_modules/displaySettingsHandler/src/displaySettingsHandler.js
```diff
@@ -0,0 +1,211 @@
...
+ * Copyright 2012 Antranig Basman
+ * Copyright 2012 Astea Solutions AD
+ * Copyright 2016 Raising the Floor - US
+ *
+ * Licensed under the New BSD license. You may not use this file except in
+ * compliance with this License.
+ *
+ * The research leading to these results has received funding from the European Union's
```
amb26: This block is no longer necessary in files created after the end of C4A

gpii/node_modules/displaySettingsHandler/src/displaySettingsHandler.js
```diff
@@ -0,0 +1,211 @@
...
+    *
+    * @return {Object) The width and height of the screen.
+    */
+windows.display.getScreenResolution = function () {
+    var dm = new windows.DEVMODEW();
+    dm.ref().fill(0);
+    dm.dmSize = windows.DEVMODEW.size;
+
+    if (c.FALSE !== windows.display.user32.EnumDisplaySettingsW(ref.NULL, c.ENUM_CURRENT_SETTINGS, dm.ref())) {
+        // note for unknown reason on win 10 the returned dmSize is 188 not expected 220
```
amb26: What is the effect of the fact that the size is 188 not 220?

gpii/node_modules/displaySettingsHandler/src/displaySettingsHandler.js
```diff
@@ -0,0 +1,211 @@
...
+
+
+/**
+    *  Sets the current display's screen resolution if possible
+    *
+    * @param {Object} The new sreen resolution width and height
+    * @return {boolean} true if successfu
+    */
+windows.display.setScreenResolution = function (newRes) {
+    if (undefined === newRes ||
```
amb26: We don't use "Yoda comparisons"

gpii/node_modules/displaySettingsHandler/src/displaySettingsHandler.js
```diff
@@ -0,0 +1,211 @@
...
+    return { width: 0, height: 0 };
+
+};
+
+
+/**
+    *  Sets the current display's screen resolution if possible
+    *
+    * @param {Object} The new sreen resolution width and height
+    * @return {boolean} true if successfu
```
amb26: The fu is great with you

gpii/node_modules/displaySettingsHandler/src/displaySettingsHandler.js
```diff
@@ -0,0 +1,211 @@
...
+    */
+windows.display.setScreenResolution = function (newRes) {
+    if (undefined === newRes ||
+        null === newRes ||
+        "object" !== typeof (newRes) ||
```
amb26: Endless and redundant block. More appropriate to throw a fatal error if we receive a newRes where typeof(newRes.width) !== "number" etc.

gpii/node_modules/displaySettingsHandler/src/displaySettingsHandler.js
```diff
@@ -0,0 +1,211 @@
...
+        undefined === newRes.width ||
+        undefined === newRes.height) {
+        return false;
+    }
+
+    var index = 0;
+    var dm = new windows.DEVMODEW();
```
amb26: This block appears 3 or more times in the code - it should become a utility function

gpii/node_modules/displaySettingsHandler/src/displaySettingsHandler.js
```diff
@@ -0,0 +1,211 @@
...
+    dm.ref().fill(0);
+    dm.dmSize = windows.DEVMODEW.size;
+
+    while (c.FALSE !== windows.display.user32.EnumDisplaySettingsW(ref.NULL, index++, dm.ref())) {
```
amb26: Eliminate Yodaism. I guess we can enforce this through a linting rule if it starts to get a foothold :)
http://eslint.org/docs/rules/yoda

gpii/node_modules/displaySettingsHandler/src/displaySettingsHandler.js
```diff
@@ -0,0 +1,211 @@
...
+};
+
+var settingName = "screen-resolution";
+
+windows.display.setImpl = function (payload) {
+    var oldRes = windows.display.getScreenResolution();
+    var newRes = oldRes;
+
+    var targetRes = payload.settings[settingName];
+    if (undefined === targetRes ||
```
amb26: See above

gpii/node_modules/displaySettingsHandler/src/displaySettingsHandler.js
```diff
@@ -0,0 +1,211 @@
...
+            if (c.FALSE !== windows.display.user32.EnumDisplaySettingsW(ref.NULL, c.ENUM_CURRENT_SETTINGS, dmCurrent.ref())) {
+                // Note: ChangeDisplaySettings has no effect if new is same as current
+                dmCurrent.dmPelsWidth = newRes.width;
+                dmCurrent.dmPelsHeight = newRes.height;
+                var lRet = windows.display.user32.ChangeDisplaySettingsW(dmCurrent.ref(), 0);
+                if (c.DISP_CHANGE_SUCCESSFUL === lRet) {
+                    return true;
+                }
+                if (c.DISP_CHANGE_RESTART === lRet) {
+                    //printf("Reset required\r\n"); // FIXME what should we do in this case? Alert user? Ignore? Roll back?
```
amb26: At the very least log an intelligible message that someone might see. Can raise a JIRA about this kind of issue in general or hook onto an existing one you can find

gpii/node_modules/displaySettingsHandler/src/displaySettingsHandler.js
```diff
@@ -0,0 +1,211 @@
...
+                    return true;
+                }
+            }
+        }
+    }
+
+    return false;
+};
+
+var settingName = "screen-resolution";
```
amb26: Undesirable "local global". What is the purpose of this setting name? Can it be usefully parameterised or changed by the user? If not, just use the literal.

gpii/node_modules/displaySettingsHandler/test/testDisplaySettingsHandler.js
```diff
@@ -0,0 +1,61 @@
+/*
+ * Windows Display Settings Handler Tests
+ *
+ * Copyright 2015 Raising the Floor - International
```
amb26: Stale, copy-pasted copyright block

gpii/node_modules/displaySettingsHandler/src/displaySettingsHandler.js
```diff
@@ -0,0 +1,211 @@
...
+windows.display.getScreenResolution = function () {
+    var dm = new windows.DEVMODEW();
+    dm.ref().fill(0);
+    dm.dmSize = windows.DEVMODEW.size;
+
+    if (c.FALSE !== windows.display.user32.EnumDisplaySettingsW(ref.NULL, c.ENUM_CURRENT_SETTINGS, dm.ref())) {
+        // note for unknown reason on win 10 the returned dmSize is 188 not expected 220
+        return { width: dm.dmPelsWidth, height: dm.dmPelsHeight };
+    }
+    return { width: 0, height: 0 };
```
amb26: i) If this is intended as a signal of an error condition, a) it should be documented, b) the value should be a globally advertised constant in your namespace, preferably frozen using fluid.freezeRecursive
ii) "dangling return" should instead be an else block

gpii/node_modules/displaySettingsHandler/test/testDisplaySettingsHandler.js
```diff
@@ -0,0 +1,61 @@
...
+require("../src/displaySettingsHandler.js");
+
+jqUnit.module("Windows Display Settings Handler Tests");
+
+jqUnit.test("Testing GetScreenResolution", function () {
+    // Test do not get the error return
+    jqUnit.expect(1);
+    var errorRes = { width: 0, height: 0};
+    var screenRes = gpii.windows.display.getScreenResolution();
+    jqUnit.assertDeepNeq("Unexpected screen resolution", errorRes, screenRes);
```
amb26: Testing for inequality is a dangerous practice and can lead to test cases which rot silently.
Also, the purpose of this test is a bit odd. It seems that the error condition doesn't really arise through our error, but through some feature of the system itself - e.g. that it has no display attached, etc. I think it would be more reasonable to either i) test that we get a record with two positive integers in the right place, and/or change the error return to not resemble this structure (e.g. return null or undefined)

gpii/node_modules/displaySettingsHandler/test/testDisplaySettingsHandler.js
```diff
@@ -0,0 +1,61 @@
...
+    var targetRes = { width: 800, height: 600 };
+    jqUnit.assertTrue("We can call to setScreenResolution, it returns 'true'", gpii.windows.display.setScreenResolution(targetRes));
+    var newRes = gpii.windows.display.getScreenResolution();
+    jqUnit.assertDeepEq("New resolution is set", targetRes, newRes);
+
+    // Restore old resolution
+    jqUnit.expect(2);
+    jqUnit.assertTrue("Error resetting old resolution", gpii.windows.display.setScreenResolution(oldRes));
+    var restoredRes = gpii.windows.display.getScreenResolution();
+    jqUnit.assertDeepEq("Old resolutions does not appear to be restored", oldRes, restoredRes);
```
amb26: Message in the assertion should be positively worded.

gpii/node_modules/displaySettingsHandler/test/testDisplaySettingsHandler.js
```diff
@@ -0,0 +1,61 @@
...
+    // Not such a good unit test as depends on available modes
+    jqUnit.expect(2);
+    var targetRes = { width: 800, height: 600 };
+    jqUnit.assertTrue("We can call to setScreenResolution, it returns 'true'", gpii.windows.display.setScreenResolution(targetRes));
+    var newRes = gpii.windows.display.getScreenResolution();
+    jqUnit.assertDeepEq("New resolution is set", targetRes, newRes);
+
+    // Restore old resolution
+    jqUnit.expect(2);
+    jqUnit.assertTrue("Error resetting old resolution", gpii.windows.display.setScreenResolution(oldRes));
```
amb26: Should be positively worded

gpii/node_modules/displaySettingsHandler/test/testDisplaySettingsHandler.js
```diff
@@ -0,0 +1,61 @@
...
+    var restoredRes = gpii.windows.display.getScreenResolution();
+    jqUnit.assertDeepEq("Old resolutions does not appear to be restored", oldRes, restoredRes);
+
+    // test can't change to invalid resolution
+    jqUnit.expect(2);
+    var badRes = { width: 123, height: 123 };
```
amb26: Why might this be an invalid resolution? I'm sure I've seen the resolution of VM windows set to figures similar to this. Negative numbers might be a more sound test.

gpii/node_modules/displaySettingsHandler/src/displaySettingsHandler.js
javihernandez: Well, I haven't experienced any problems on Win10, but besides the regular "Struct size differs" errors (where we should explode) or receiving wrong data due to unaligned bytes in the DEVMODE structure (where will drive us to unexpected behaviors of the system by reading values from the wrong memory position) I don't see any other side effects. Since the DEVMODE structure is a crazy one (https://msdn.microsoft.com/en-us/library/dd183565.aspx), its size may differ depending on the version of windows, the device you're dealing with (display, printer), etc, I think that I can test this code in different versions of Windows and see what happens in different versions, and whether this difference in size represents a real problem. Other than that, any ideas?

gpii/node_modules/displaySettingsHandler/src/displaySettingsHandler.js
javihernandez: Well, I'd say that we should raise an error if we can't get the screen resolution. I've changed this to raise an error if we can't get the screen resolution so the system can take this into account. Let me know if this makes sense to you.

sgithens:
This just needs another review from @amb26  to make sure it addressed his comments right? 

@javihernandez : you're done with any changes right now correct?

javihernandez:
> This just needs another review from @amb26 to make sure it addressed his comments right?
> 
> @javihernandez : you're done with any changes right now correct?

@sgithens Yes, I'm done, but for whatever reason I don't know why Github is still showing comments for code that has been updated.

amb26:
This only passes CI because none of the functionality in it is exercised. Please could you i) merge this up with master, ii) merge up https://github.com/GPII/universal/pull/465 with its master, and iii) bind the package.json reference in this pull to the universal reference that is generated in ii) in order to allow the CI to test it fully

javihernandez:
@amb26 ready again.

amb26:
Great work, @javihernandez ! This is now all merged in. Good work on a really important and long-requested feature.

PR #83, opened 2016-08-08
GPII-581: Investigate means for detecting the time point of application of dynamic settings (esp. for Windows SPI settings)

stegru:
Calls to SPI will return a promise, which will resolve when the setting has really been applied (SPI_SETHIGHCONTRAST is the only async one).


amb26:
This is just brilliant - not least, because it seems to actually work! : P This might be the first time in 4 years we've been able to run the acceptance tests on Windows without leaving the desktop corrupted. Really good work, Steve! Some minor cleanup comments are scattered around

gpii/node_modules/WindowsUtilities/WindowsUtilities.js
```diff
@@ -19,10 +19,17 @@
 "use strict";
 
 var ffi = require("ffi");
-var fluid = require("universal");
+var gpii, windows, fluid;
 
-var gpii = fluid.registerNamespace("gpii");
-var windows = fluid.registerNamespace("gpii.windows");
+if (process.env.GPII_CHILD_PROCESS) {
```
amb26: The approach of identifying the child process via an environment variable is ok, but we should avoid special-casing it in the core code. If there is some material in this file which we can usefully include i) in a lightweight child process, and ii) which has fewer dependencies than the rest of the file, we should highlight this by splitting off the material into a separate file with an appropriate comment. For example, something like "WindowsUtilitiesLight.js" etc.

amb26: In general we should be able to load fluid without loading universal - but due to a fault in our current module strategy, we can't. This is something that we should be able to remedy once we make the move over to node 6 and npm 3.

gpii/node_modules/spiSettingsHandler/src/SpiHighContrastCompletion.js
```diff
@@ -0,0 +1,122 @@
...
+                                waited = true;
+                                outOfTime = true;
+                                break;
+                            default:
+                                // failure
+                                break;
+                            }
+                        }
+
+                        if (!waited) {
```
amb26: What are some reasons that we may have failed to be able to wait for the process?

gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
```diff
@@ -82,6 +82,57 @@ gpii.windows.spi.getCurrentSettings = function (payload) {
...
+    var options = {
+        // Let the child process know it's a child process
+        env: {
+            GPII_CHILD_PROCESS: 1,
+            GPII_SETHC_EXE: exeFile || "sethc.exe"
+        }
+    };
+
+    // If debugging, remove the debugger listeners
+    if (typeof v8debug === "object") {
```
amb26: This is pretty interesting. Was this code lifted from somewhere, and if so, where? What is the result if we don't remove the debugging listeners?

gpii/node_modules/spiSettingsHandler/test/testSpiSettingsHandler.js
```diff
@@ -20,16 +20,55 @@ https://github.com/GPII/universal/blob/master/LICENSE.txt
...
-jqUnit.module("SpiSettingsHandler Module");
+require("../src/SpiSettingsHandler.js");
+
+var waitExe = "fake-sethc.exe";
+var waitExePath;
+
+jqUnit.module("SpiSettingsHandler Module", {
+    teardown: function () {
+        if (waitExePath !== null) {
+            shelljs.rm(waitExePath);
```
amb26: Very good

gpii/node_modules/spiSettingsHandler/test/testSpiSettingsHandler.js
```diff
@@ -20,16 +20,55 @@ https://github.com/GPII/universal/blob/master/LICENSE.txt
...
+        }
+    }
+});
+
+
+jqUnit.asyncTest("SpiSettingsHandler HighContrast Completion", function () {
+
+    // Take a copy of the built-in "waitfor" command, to ensure a unique process name
+    waitExePath = path.join(process.env.TEMP, waitExe);
+    shelljs.cp(path.join(process.env.SystemRoot, "/System32/waitfor.exe"), waitExePath);
```
amb26: Amazing.... and it even exists in Windows XP : P

gpii/node_modules/spiSettingsHandler/src/SpiHighContrastCompletion.js
```diff
@@ -0,0 +1,122 @@
+/*
+ * SPI High-contrast completion. Waits for the completion of the SPI_SETHIGHCONTRAST call.
```
amb26: Should expand this comment a little to identify the strategy used in detail - e.g. the special significance of sethc.exe, anything we have learned about it, and the particular API strategy used to interact with it

gpii/node_modules/spiSettingsHandler/src/SpiHighContrastCompletion.js
```diff
@@ -0,0 +1,122 @@
...
+                            windows.kernel32.Sleep(800);
+                        }
+                    } finally {
+                        if (hProcess && hProcess !== ref.NULL) {
+                            windows.kernel32.CloseHandle(hProcess);
+                            hProcess = null;
+                        }
+                    }
+
+                    // There may be more processes running, but the current snap-shot may be stale.
```
amb26: Not sure I grasp this comment - what is the condition we fall through here?

gpii/node_modules/spiSettingsHandler/src/SpiHighContrastCompletion.js
```diff
@@ -0,0 +1,122 @@
...
+    var found = false;
+
+    var timer = setTimeout(function () {
+        outOfTime = true;
+    }, timeout);
+
+    // Loop until no matching processes are found.
+    do {
+        found = false;
+        var hSnapShot = windows.kernel32.CreateToolhelp32Snapshot(windows.API_constants.TH32CS_SNAPPROCESS, null);
```
amb26: Split up this complex function with deeply nested control structures into (perhaps) 3 separate functions, with less (or hopefully zero) reliance on shared top-level function scope "globals" e.g. outOfTime, found, hSnapshot. It looks like there's a role for a function at this level within the "do" and then another one inside while (hRes)

gpii/node_modules/WindowsUtilities/WindowsUtilities.js
amb26: Perhaps WindowsUtilitiesCore.js

gpii/node_modules/spiSettingsHandler/src/SpiHighContrastCompletion.js
```diff
@@ -0,0 +1,122 @@
...
+ * Waits for a process to terminate, returning when there are no matching processes.
+ *
+ * A snapshot of the running processes is taken, and if a matching process is found try to
+ * open it and wait for it to close.
+ *
+ *
+ * @param filename The executable.
+ * @param timeout The timeout (ms).
+ */
+windows.waitForProcessTermination = function (filename, timeout) {
```
amb26: Given this whole algorithm is generic, it should go into its own implementation file

gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
```diff
@@ -82,6 +82,57 @@ gpii.windows.spi.getCurrentSettings = function (payload) {
...
+            }
+        }
+    }
+
+    var child = cp.fork(__dirname + "/SpiHighContrastCompletion", options);
+
+    var promise = fluid.promise();
+
+    child.on("exit", function (code) {
+        promise.resolve(code);
```
amb26: If the code is non-zero this should be a reject

gpii/node_modules/WindowsUtilities/WindowsUtilities.js
```diff
@@ -19,10 +19,17 @@
...
+    // If this is a child process then don't load universal, or anything else that is not required.
+    fluid = gpii = windows = {};
```
amb26: We should add a comment here to reference https://issues.gpii.net/browse/GPII-1912 to remind us to remove this strategy when we can. Right now we have a bunch of console.log statements in the child process that can't be configured away. Ideally we'd like these to be fluid.log calls, but since we don't yet support a reasonable scheme for the integrator to configure the Infusion logging level, this isn't particularly urgent.

gpii/node_modules/spiSettingsHandler/src/SpiHighContrastCompletion.js
stegru: I don't know, however I don't think it matters why as there's nothing much that can be done to fix it. Instead, Sleep is called just after this.

What we do know at that point is the process still exists, so the line just after this snippet pauses the process if there was an error, and the loop will continue doing this until the process is no longer there.

gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
stegru: It happened to me. The --debug/--debug-brk arguments cause node to listen on a certain tcp port - however that port is already being listened on by the main process. Also, --debug-brk halts node until my debugger had attached.

These arguments at removed from the call to the child-process to avoid this, I don't think they will be there in production.

gpii/node_modules/spiSettingsHandler/src/SpiHighContrastCompletion.js
stegru: This is at the end of the "I have found the process" block. At this point, the process would have been waited upon and terminated - but what if there where more than one of these processes (or, the wait failed and this one is still running)? I still want to check the task list again.

Because of the wait, the list of running processes could have changed by now, so instead of continuing to work on the last process snapshot, I break out of the enumeration to get a fresh one

gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
amb26: Ok - could you add a comment here to explain that the issue relates to the blocked port with respect to the main process. Also, why do we copy down all the other exec arguments from the main process - are there any of them in particular that we're interested in?

stegru: I don't know about the arguments. If you don't specify any, then all of the arguments that the main process has are passed on by default. My intention was to make it behave like that, but without those --debug ones.

The reason why being if the main process was passed "--make-it-work", then why not have the child process run in the same manner.

The debug issue "documented" on [stackoverflow](http://stackoverflow.com/questions/16840623/how-to-debug-node-js-child-forked-process)

stegru: Oh it looks like they had the same problem with the "cluster" API, and had a similar workaround.
https://github.com/nodejs/node-v0.x-archive/commit/43ec1b1c2e77d21c7571acd39860b9783aaf5175#diff-0faa53fc02580d5de2ebb484c41d691cR67

amb26: Ok, that's interesting. Is there possibly a route to simplifying all of this by simply execing node.exe rather than forking ourselves? I think all of the node toolchain assumes that node.exe is on PATH anyway.

stegru: Yes that would be simpler, and I can just execute argv[0]

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -0,0 +1,151 @@
...
+ * The promise will reject if the process is still running after the timeout.
+ *
+ * @param filename The executable.
+ * @param timeout Approximate milliseconds to wait, or null for infinite.
+ * @return {Promise} The promise will resolve when there are no matching processes running.
+ */
+gpii.windows.waitForProcessTermination = function (filename, timeout) {
+    var timeoutTime = null;
+    if (timeout) {
+        timeoutTime = (new Date).getTime() + timeout;
```
amb26: This can just be Date.now()

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -0,0 +1,151 @@
...
+};
+
+/**
+ * Waits for a process to terminate, returning a promise which will resolve when there are no matching processes
+ * running. If there are no processes running when the function is called, the promise will already be resolved.
+ * The promise will reject if the process is still running after the timeout.
+ *
+ * @param filename The executable.
+ * @param timeout Approximate milliseconds to wait, or null for infinite.
+ * @return {Promise} The promise will resolve when there are no matching processes running.
```
amb26: Also state that it will reject on timeout

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -0,0 +1,151 @@
...
+    // Continuously check for a matching process, until there is none.
+    var checkProcess = function () {
+        // Get the pid of a matching process.
+        var pid = windows.findProcessByName(filename);
+
+        if (pid === null) {
+            // No matching process.
+            promise.resolve();
+        } else {
+            if (!timeout || ((new Date).getTime() < timeoutTime)) {
```
amb26: Same again

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -0,0 +1,151 @@
...
+/**
+ * Waits for a process to terminate, returning a promise which will resolve when there are no matching processes
+ * running. If there are no processes running when the function is called, the promise will already be resolved.
+ * The promise will reject if the process is still running after the timeout.
+ *
+ * @param filename The executable.
+ * @param timeout Approximate milliseconds to wait, or null for infinite.
+ * @return {Promise} The promise will resolve when there are no matching processes running.
+ */
+gpii.windows.waitForProcessTermination = function (filename, timeout) {
```
amb26: Turn the 2nd argument into an "options" structure which can not only incorporate timeout, but also the interval with room to expand for any other directives

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -0,0 +1,151 @@
...
+        // Get the pid of a matching process.
+        var pid = windows.findProcessByName(filename);
+
+        if (pid === null) {
+            // No matching process.
+            promise.resolve();
+        } else {
+            if (!timeout || ((new Date).getTime() < timeoutTime)) {
+                // It's still there - keep checking.
+                setTimeout(checkProcess, 500);
```
amb26: Stick this hard-coded interval into the options strucure

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -0,0 +1,151 @@
...
+            } else {
+                // Ran out of time.
+                promise.reject();
```
amb26: Our general policy for rejection objects is for them to look something like this:
{
    isError: true,
   message: "Timed out waiting for process xxxx to terminate after yyyy ms"
}

gpii/node_modules/processHandling/test/testProcessHandling.js
```diff
@@ -0,0 +1,136 @@
...
+        shelljs.cp(path.join(process.env.SystemRoot, "/System32/waitfor.exe"), waitExePath);
+    },
+    teardown: function () {
+        if (waitExePath !== null) {
+            shelljs.rm(waitExePath);
+        }
+    }
+});
+
+gpii.tests.waitForTask = function (task, callback, maxCount) {
```
amb26: This can now be refactored away to use the waitForProcessTermination utility you just wrote

gpii/node_modules/WindowsUtilities/WindowsUtilities.js
```diff
@@ -89,6 +89,14 @@ windows.kernel32 = ffi.Library("kernel32", {
     // https://msdn.microsoft.com/en-us/library/windows/desktop/ms684836%28v=vs.85%29.aspx
     "Process32Next": [
         t.BOOL, [t.HANDLE, "pointer"]
+    ],
+    // https://msdn.microsoft.com/en-us/library/windows/desktop/ms687032.aspx
+    "WaitForSingleObject": [
```
amb26: We should eliminate this definition since it is unused. Whilst it is nice to have extra definitions lying around in case of use, in practice code which has no test cases will silently rott......

gpii/node_modules/WindowsUtilities/WindowsUtilities.js
```diff
@@ -111,6 +119,11 @@ windows.API_constants = {
     },
     // https://msdn.microsoft.com/en-us/library/windows/desktop/ms684880%28v=vs.85%29.aspx
     PROCESS_TERMINATE: 0x0001,
+    SYNCHRONIZE: 0x00100000,
+    // https://msdn.microsoft.com/en-us/library/windows/desktop/ms687032.aspx
+    WAIT_OBJECT_0: 0,
```
amb26: Similarly

gpii/node_modules/processHandling/test/testProcessHandling.js
stegru: waitForTask waits for a task to _start_ (or more accurately, waits for task() to return true). A comment will be added.

For both calls to this, the 'task' argument is a function that already uses the new findProcessByName function.

stegru:
Fixes complete.

gpii/node_modules/processHandling/test/testProcessHandling.js
amb26: Ok - the core algorithms are essentially identical though so I think this is a good candidate for consolidation. "waitForProcessStateChange" with an entry in the options structure "start: true/false"? As a courtesy, we could retain waitForProcessTermination and waitForProcessStart as wrapper functions.

gloob:
ok to test

gpii/node_modules/processHandling/test/testProcessHandling.js
stegru: done

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -0,0 +1,214 @@
...
+
+    var promise = fluid.promise();
+
+    // Continuously check for a matching process, until there is none.
+    var checkProcess = function () {
+        // Get the pid of a matching process.
+        var running = windows.findProcessByName(filename) !== null;
+        var changed = false;
+
+        if (options.start === null) {
```
amb26: I think the "null" option is likely to lead to problems and potentially even race conditions if the system state changes in between some previous check for whether a process is running and the execution of this function. We should just stick with supporting plain "true" and "false" - this would also let us get rid of the oddity on line 144 where we modify the options structure. Whilst you've taken care on line 128 to ensure that we've cloned the user's options (at least shallowly) our general policy is to consider that "options" structures are immutable after the initial merging process has completed.

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -0,0 +1,214 @@
...
+ * @param filename The executable.
+ * @param options The options.
+ * @param options.timeout Approximate milliseconds to wait, or null for infinite.
+ * @param options.pollDelay How long to wait, in milliseconds, between checking for the process.
+ * @return {promise} The promise will resolve when there is a matching processes running, or will reject if a matching
+ *  process is still not running after the timeout.
+ */
+gpii.windows.waitForProcessStart = function (filename, options) {
+    options = options || {};
+    options.start = true;
```
amb26: The user's input structure is modified here, which we generally discourage - safer and more direct is
function (filename, userOptions) {
    var options = fluid.extend(true, {start: true}, userOptions});
etc.

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -0,0 +1,214 @@
...
+ *
+ * @param filename The executable.
+ * @param options The options.
+ * @param options.timeout Approximate milliseconds to wait, or null for infinite.
+ * @param options.pollDelay How long to wait, in milliseconds, between checking for the process.
+ * @return {promise} The promise will resolve when there are no matching processes running, or will reject if a matching
+ *  process is still running after the timeout.
+ */
+gpii.windows.waitForProcessTermination = function (filename, options) {
+    options = options || {};
```
amb26: See below

gpii/node_modules/processHandling/processHandling.js
amb26: Also - this branch is untested!

gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
```diff
@@ -82,6 +83,45 @@ gpii.windows.spi.getCurrentSettings = function (payload) {
...
+ * @return {promise} A promise that resolves when the last SPI call is complete.
+ */
+gpii.windows.spi.waitForSpi = function (action) {
+    if (action && action !== gpii.windows.actionConstants.SPI_SETHIGHCONTRAST) {
+        // The setting has been applied as soon as SPI returns.
+        return fluid.promise().resolve();
+    }
+
+    /*
+     * Experimentation has uncovered an undocumented occurrence when the high-contrast is changed
```
amb26: Really great comment

gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
```diff
@@ -82,6 +83,45 @@ gpii.windows.spi.getCurrentSettings = function (payload) {
...
+     * via the SPI call, where Windows executes %WINDIR%\system32\sethc.exe to perform the actual
+     * work of changing the contrast. Investigations have shown that this executable may be
+     * called with parameters "10", "11", "100", or "101" (numbers < 100 turn HC off).
+     *
+     * It has been noticed that the work behind setting SPI_SETHIGHCONTRAST is complete when
+     * this process terminates, so this script checks for the existence of this process and
+     * does not return until the has ended.
+     */
+    return gpii.windows.waitForProcessTermination("sethc.exe");
```
amb26: Just for being "clean under the fingernails" do use an else block here rather than a fallthrough - cheers!

gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
```diff
@@ -82,6 +83,45 @@ gpii.windows.spi.getCurrentSettings = function (payload) {
...
+
+/**
+ * Waits for a previous call to SPI to complete. This is to work-around the high-contrast change
+ * taking some time to complete, and to prevent other calls in the meantime.
+ *
+ * @param action The action parameter of the SystemParametersInfo call.
+ * @return {promise} A promise that resolves when the last SPI call is complete.
+ */
+gpii.windows.spi.waitForSpi = function (action) {
+    if (action && action !== gpii.windows.actionConstants.SPI_SETHIGHCONTRAST) {
```
amb26: No need for the first branch here, equality checks are always safe in JS. In general, we favour positive conditions - 

if (action === gpii.windows.actionConstants.SPI_SETHIGHCONTRAST) {
etc.

amb26:
Thanks @ste - looking really great and clean - just a few minor tweaks to the logic - after you've cleaned up, could you i) merge this up with current master and then ii) add the "ok to test" comment so that we can see whether this passes CI? Somehow gloob's previous comment didn't seem to trigger a rebuild...

gpii/node_modules/processHandling/processHandling.js
stegru: I know - and I can't even think of a use case for it.

stegru:
ok to test

amb26:
Thanks for the brilliant work on this vital issue, @stegru - as well as your thoroughness in tracking down an interesting and stable route for solving the issue, you've shown a brilliant commitment during the review process to polishing the original solution until it is simple, clear and minimal. Really great!

colinbdclark:
Congratulations on getting this pull request merged into the GPII, @stegru! This was a crucial fix and took a lot of effort to prepare; thank you so much! 👍

PR #84, opened 2016-08-13
GPII-1898: Update to node 6

javihernandez:
This isn't ready yet until these two issues are solved:
* https://issues.gpii.net/browse/GPII-1929 **SOLVED**
* https://issues.fluidproject.org/browse/FLUID-5940 **SOLVED**

Can be tested by applying the following patch: **NO LONGER NEEDED**
```
diff --git a/package.json b/package.json
index 3a8ea2b..7df85a8 100644
--- a/package.json
+++ b/package.json
@@ -10,7 +10,9 @@
         "ref": "1",
         "ref-struct": "1",
         "ref-array": "1.1.2",
-        "universal": "gpii/universal#98561d07c0b49d4d565be9027adebf36c1d1fe82"
+        "universal": "gpii/universal#98561d07c0b49d4d565be9027adebf36c1d1fe82",
+        "nan": "amatas/nan#bbe2ba05648320ce19646e2ceab9c840cd59583e",
+        "infusion": "javihernandez/infusion#77c3abb599648635fad3870bce8702411b677ca7"
     },
     "devDependencies": {
         "grunt": "1.0.1",
```
I'm intentionally bumping the minimum node version required to 6.* because we need npm 3 to avoid [this problem](https://github.com/GPII/windows/pull/76#issuecomment-235427320).


gloob:
ok to test

colinbdclark:
ok to test

amb26:
@gloob - there's a fix in Infusion core now for FLUID-5940 but https://issues.gpii.net/browse/GPII-1994 is now a new blocker assigned to @kaspermarkus

javihernandez:
I've updated this with current master to avoid conflicts.

The current status of the blockers we have is the following:
* https://issues.fluidproject.org/browse/FLUID-5940 has been resolved but still waiting for the resolution of https://issues.gpii.net/browse/GPII-1994 (which is almost there)
* Looks like the issue we're having with nan goes beyond (see https://github.com/nodejs/nan/pull/601) and the resolution may take some time. We're discussing the alternatives https://issues.gpii.net/browse/GPII-1929.

BTW, I've reverted the minimum required version of node in universal and linux BUT since we're having this particular problem with windows, I recommend to keep the node version requirement to 6.x so users don't need to manually update npm to get it working with node 4.x.

javihernandez:
I've updated the reference to universal and added nan as dependency. I've also created an issue (https://issues.gpii.net/browse/GPII-2037) to remove our dependency with nan when the issue we have is fixed upstream.

PR #86, opened 2016-08-30
GPII 1540: AT persists after sleep/lock/UAC

stegru:
This fix adds the ability to kill applications an less aggressive way by sending WM_QUIT to their windows, allowing the application to perform its own clean-up, and unregister itself as AT.

Also: https://github.com/GPII/universal/pull/469


gpii/node_modules/processHandling/test/test-window.c
```diff
@@ -0,0 +1,79 @@
...
+ */
+
+
+#include <Windows.h>
+#include <stdio.h>
+#pragma comment (lib, "User32.lib")
+
+
+/* Creates a window and message pump, returns when the window closes. */
+int window() {
```
amb26: Miraculous and brilliant!

gpii/node_modules/processHandling/test/test-window.c
```diff
@@ -0,0 +1,79 @@
+/*
+ * test-window.c: Simple application that creates a window, used to test closeProcessByName.
+ *
+ * How to build:
```
amb26: We need to add a build for this app to our standard npm install build for Windows so that our tests work out of the box

amb26: Without this, we can't get this pull past CI

stegru:
ok to test

PR #90, opened 2016-09-26
GPII-2001: Unit tests get stalled at some point during its execution

stegru:
Worked around this problem by making the SPI_SETNONCLIENTMETRICS SPI call in a separate process. This will prevent the main process from being blocked.


stegru:
ok to test

colinbdclark:
ok to test

PR #98, opened 2016-10-24
GPII-2073: Persona Settings return on windows lock screen

stegru:
Starting and stopping the built-in AT needs a bit more consideration than usual, as they like to register themselves so the Windows can start/stop them when the desktop changes. This PR ensures the AT is disabled upon logout, so it wont re-appear at the lock screen. It also uses the executable as registered in the Ease of access centre (I'm not confident this is desirable, so an override is available).


stegru:
ok to test

javihernandez:
ok to test

gpii/node_modules/registeredAt/registeredAt.js
```diff
@@ -0,0 +1,248 @@
...
+ *   some other 3rd party name.
+ * @param {Boolean} enable true to enable the application.
+ * @param {{}} options Options
+ * @param {Boolean} options.configOnly true to only (de)register the application and not to start or stop it.
+ * @param {String} options.startCommand Overrides the command used to start the AT application.
+ * @param {String} options.processName Overrides the executable to terminate when disabling.
+ * @param {String} options.utilman Overrides the command for utilman.exe
+ * @return {*}
+ */
+gpii.windows.enableRegisteredAt = function (name, enable, options) {
```
amb26: I think enableRegisteredAT is a clearer name, regardless of what Microsoft feels about such things :)

gpii/node_modules/registeredAt/registeredAt.js
```diff
@@ -0,0 +1,248 @@
...
+
+    // The AT will have enabled itself when it starts, so only update the registry when disabling.
+    if (enable && !options.configOnly) {
+        promise.resolve();
+    } else {
+        // On Windows 7/8, if there's more than one action value in the registry then it will display the Ease of Access
+        // Centre. (This could be because when the first AT is loaded, it then invokes yet another instance of utilman,
+        // which will work on the next value pulling the rug from under the first instance).
+        // To work around this, wait until it has been cleared.
+        gpii.windows.whilePendingAt().then(function () {
```
amb26: Do normalise case to whilePendingAT etc.

gpii/node_modules/registeredAt/registeredAt.js
```diff
@@ -0,0 +1,248 @@
...
+ * The temporary key should have the same name as the registered application name, such as [magnifierpane, osk, or
+ * Narrator].
+ * The value of the key is a DWORD set to 0x0003 when it is starting, or 0x0002 when the application is exiting."
+ *
+ * Sending this key sequence is not necessary for GPII, as either the AT will do that itself when it starts, or GPII
+ * will execute utilman.exe which performs the necessary work.
+ */
+
+// The key where the temporary values are kept.
+var accessibilityTempKey = {
```
amb26: Put these into publically namespaced constants in case they are useful somewhere - gpii.windows.registeredAT.* or so

gpii/node_modules/registeredAt/registeredAt.js
```diff
@@ -0,0 +1,248 @@
...
+/**
+ * Enable or disable a registered AT application.
+ *
+ * While an application is enabled, Windows will automatically start and stop the application when the user switches
+ * desktops. Disabling will stop this from happening.
+ *
+ * @param {String} name The registered name of the application. This can be "magnifierpane", "osk", "Narrator", or
+ *   some other 3rd party name.
+ * @param {Boolean} enable true to enable the application.
+ * @param {{}} options Options
```
amb26: {Object} is clearer

gpii/node_modules/registeredAt/registeredAt.js
```diff
@@ -0,0 +1,248 @@
...
+        // On Windows 7/8, if there's more than one action value in the registry then it will display the Ease of Access
+        // Centre. (This could be because when the first AT is loaded, it then invokes yet another instance of utilman,
+        // which will work on the next value pulling the rug from under the first instance).
+        // To work around this, wait until it has been cleared.
+        gpii.windows.whilePendingAt().then(function () {
+            var targetValue = enable ? enableAt : disableAt;
+            gpii.windows.writeRegistryKey(accessibilityTempKey.baseKey, accessibilityTempKey.path, name, targetValue,
+                "REG_DWORD");
+            // Invoke utilman to trigger the de-registration.
+            gpii.windows.nativeExec(options.utilman);
```
amb26: Well worth checking for a non-zero return code and sending it to promise.reject() here

gpii/node_modules/registeredAt/registeredAt.js
```diff
@@ -0,0 +1,248 @@
...
+
+    var timeTaken = 0;
+    var promise = fluid.promise();
+
+    var checkActions = function () {
+        // Scan the AccessibilityTemp key for action values.
+        var found = false;
+        gpii.windows.enumRegistryValues(accessibilityTempKey.baseKey, accessibilityTempKey.path,
+            function (name, data) {
+                if ((data === enableAt) || (data === enableAt)) {
```
amb26: Typo here? Perhaps meant disableAT for 2nd

gpii/node_modules/registeredAt/registeredAt.js
```diff
@@ -0,0 +1,248 @@
...
+        });
+    }
+
+    return promise;
+};
+
+/**
+ * Waits while there are pending actions for utilman.
+ *
+ * @return {promise} Returns a promise, whose value is true upon timeout, resolving when there are no pending actions.
```
amb26: Some more informative return codes might help the user and also give more room for expansion - for example, resolving to "timeout" on timeout, etc.

gpii/node_modules/registeredAt/registeredAt.js
```diff
@@ -0,0 +1,248 @@
...
+    checkActions();
+    return promise;
+};
+
+/**
+ * Starts a registered AT application.
+ *
+ * @param {String} name The registered name of the application. This can be "magnifierpane", "osk", "Narrator", or
+ *   some other 3rd party name.
+ * @param {{}} [options] Options
```
amb26: {Object}

gpii/node_modules/registeredAt/test/testRegisteredAt.js
```diff
@@ -0,0 +1,198 @@
...
+        gpii.windows.killProcessByName(fakeUtilman);
+        shelljs.rm(fakeUtilmanPath);
+    });
+
+    var options = {
+        configOnly: true,
+        utilman: fakeUtilmanPath + " /T 20 gpiiFakeUtilman"
+    };
+
+    var enable = function () {
```
amb26: Fold these near-identical functions into one

gpii/node_modules/registrySettingsHandler/src/RegistrySettingsHandler.js
```diff
@@ -229,6 +235,84 @@ windows.deleteRegistryKey = function (baseKey, path) {
...
+        var nameLength = new Buffer(4);
+        var typeHolder = new Buffer(4);
+        var nameHolder = new Buffer(maxLength);
+        var dataSizeHolder = new Buffer(maxLength);
+
+        for (var index = 0; ; index++) {
+            // Get the name, type, and data length of the next value.
+            nameLength.writeUInt32LE(Number(maxLength), 0);
+            var code2 = advapi32.RegEnumValueW(keyHolder.readUInt32LE(0), index, nameHolder, nameLength, NULL, typeHolder, NULL, dataSizeHolder);
+            if (code2 === 2) { // FILE_NOT_FOUND
```
amb26: Turn these into some global symbolic constants

gpii/node_modules/registrySettingsHandler/src/RegistrySettingsHandler.js
```diff
@@ -229,6 +235,84 @@ windows.deleteRegistryKey = function (baseKey, path) {
     cr(code);
 };
 
+/**
+ * Enumerates the values in a key.
+ *
+ * @param {String} baseKey The registry base key.
+ * @param {String} path The key path.
+ * @param {function (name, data, type)} callback A call-back for each value found, returning true to break the enumeration.
```
amb26: {Function} - also list the types and meaning of the arguments

gpii/node_modules/registeredAt/registeredAt.js
```diff
@@ -0,0 +1,248 @@
...
+ * @param {String} options.startCommand Overrides the command used to start the AT application.
+ */
+gpii.windows.startRegisteredAt = function (name, options) {
```
amb26: Sending in a JSON structure representing the "AT information database" as a further argument would allow the removal of the unsightly monkey-patching in the tests

gpii/node_modules/registeredAt/test/testRegisteredAt.js
```diff
@@ -0,0 +1,198 @@
...
+    // Take a copy of the built-in "waitfor" command, to ensure a unique process name.
+    shelljs.cp(path.join(process.env.SystemRoot, "/System32/waitfor.exe"), testData.StartExe);
+    // Remove it at the end
+    teardowns.push(function () {
+        gpii.windows.killProcessByName(testData.ATExe);
+        shelljs.rm(testData.StartExe);
+    });
+
+    // Make getAtInformation return the test data. Override the function instead of dealing with the registry.
+    var getAtInformationReal = gpii.windows.getAtInformation;
```
amb26: Refactor start/stopRegisteredAT as suggested so we can clear up this awkward monkey-patching - such things should always be a last resort when dealing with recalcitrant 3rd-party code :)

gpii/node_modules/registrySettingsHandler/src/RegistrySettingsHandler.js
stegru: No longer relevant, due to your next comment.

stegru:
Ready for re-review

gpii/node_modules/registeredAT/registeredAT.js
```diff
@@ -0,0 +1,279 @@
...
+    registeredATKey: {
+        baseKey: "HKEY_LOCAL_MACHINE",
+        path: "SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion\\Accessibility\\ATs"
+    }
+});
+
+// The AccessibilityTemp values.
+// The known values of this setting are: 0 (removed), 1 (installed), 2 (removal required), 3 (install required).
+// Values 0 and 1 are informational, which are set (by utilman) as a result of 2 and 3 to prevent a repeat action.
+var disableAT = 2,
```
amb26: Make these public constants in windows.API_constants

gpii/node_modules/registeredAT/registeredAT.js
```diff
@@ -0,0 +1,279 @@
...
+                }
+            });
+        });
+    }
+
+    return promise;
+};
+
+/**
+ * Waits while there are pending actions for utilman.
```
amb26: Could you flesh out this comment a little more to explain what strategy is used to wait for the actions? It looks like we poll until a particular registry key has changed state?

gpii/node_modules/registeredAT/registeredAT.js
```diff
@@ -0,0 +1,279 @@
...
+    var promise = fluid.promise();
+
+    var checkActions = function () {
+        // Scan the AccessibilityTemp key for action values.
+        var values = gpii.windows.enumRegistryValues(gpii.windows.registeredAT.accessibilityTempKey.baseKey,
+            gpii.windows.registeredAT.accessibilityTempKey.path);
+
+        var found = false;
+        for (var name in values) {
+            if (values.hasOwnProperty(name)) {
```
amb26: slightly simpler - var data = fluid.get(values, [name, "data"]);

gpii/node_modules/registeredAT/registeredAT.js
```diff
@@ -0,0 +1,279 @@
...
+
+/**
+ * Waits while there are pending actions for utilman.
+ * @param [options] Options
+ * @param options.pollDelay How often to check for changes.
+ * @param options.timeout Milliseconds to timeout after.
+ * @return {promise} Returns a promise, resolving when there are no pending actions. The value is "timeout" on timeout.
+ */
+gpii.windows.whilePendingAT = function (options) {
+    var defaultOptions = {
```
amb26: I think we've now accumulated a critical mass of these functions with the same workflow - e.g. https://github.com/GPII/windows/blob/master/gpii/node_modules/processHandling/processHandling.js#L119 is essentially identical. Could you refactor these into a generic utility named something like "gpii.waitForCondition", accepting a generic options block like this, which then also receives the check function and then manages the workflow of i)  merging options, ii) repeatedly polling, iii) resolving to a promise - cheers! 

Also have a scour around and see if you can find any other candidates for being expressed in terms of this.

amb26: I think it can go in WindowsUtilities for now - we'll have to remember it is there and haul it up into universal someday, although that will be less hassle once we have gone for a "monorepo"

stegru:
ready for re-review

gpii/node_modules/WindowsUtilities/WindowsUtilities.js
```diff
@@ -654,5 +664,86 @@ windows.enumerateWindows = function (callback) {
...
+ *
+ * The function is called repeatedly (with a delay) until it returns a truthful value, or a period of time has elapsed.
+ *
+ * @param {Function} func The function to call which checks the condition.
+ * @param {object} [options] Options
+ * @param {*} options.conditionValue The value that the function will return to indicate the condition is met.
+ *  This is compared using '===', but the default is anything that resolves to true.
+ * @param {number} options.argument An argument to pass to the function.
+ * @param {number} options.timeout How long to wait (ms) before timing out, or null (default) for no timeout.
+ * @param {number} options.pollDelay The delay (ms) between each poll.
```
amb26: Cite default delay here

gpii/node_modules/WindowsUtilities/WindowsUtilities.js
```diff
@@ -654,5 +664,86 @@ windows.enumerateWindows = function (callback) {
...
+
+/**
+ * Waits for a condition, by polling a given function. A promise is returned, resolving when the condition is met or
+ * rejecting upon timeout. If the condition is already met, then the returned promise will be resolved.
+ *
+ * The function is called repeatedly (with a delay) until it returns a truthful value, or a period of time has elapsed.
+ *
+ * @param {Function} func The function to call which checks the condition.
+ * @param {object} [options] Options
+ * @param {*} options.conditionValue The value that the function will return to indicate the condition is met.
```
amb26: We tend to write {Any} for these types.

gpii/node_modules/WindowsUtilities/WindowsUtilities.js
```diff
@@ -654,5 +664,86 @@ windows.enumerateWindows = function (callback) {
...
+};
+
+/**
+ * Waits for a condition, by polling a given function. A promise is returned, resolving when the condition is met or
+ * rejecting upon timeout. If the condition is already met, then the returned promise will be resolved.
+ *
+ * The function is called repeatedly (with a delay) until it returns a truthful value, or a period of time has elapsed.
+ *
+ * @param {Function} func The function to call which checks the condition.
+ * @param {object} [options] Options
```
amb26: For all of these types capitalise as {Object}, {Number} etc

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -126,38 +137,19 @@ gpii.windows.waitForProcessState = function (filename, options) {
...
+        error: {
+            isError: true,
+            message: "Timed out waiting for process " + filename +
+            " to " + (options.start ? "start" : "terminate") + " after " + options.timeout + "ms"
         }
     };
 
-    checkProcess(promise);
-    return promise;
+    return gpii.windows.waitForCondition(gpii.windows.isProcessRunning, waitOptions);
```
amb26: Good job

gpii/node_modules/registeredAT/registeredAT.js
```diff
@@ -0,0 +1,271 @@
...
+
+/**
+ * Stops a registered AT application.
+ *
+ * @param {String} name The registered name of the application. This can be "magnifierpane", "osk", "Narrator", or
+ *   some other 3rd party name.
+ * @param {{}} [options] Options
+ * @param {String} options.processName Overrides the executable name to terminate.
+ * @param {Object} options.atInfo Overrides the AT information taken from the registry.
+ * @return {promise} A promise that will resolve when the application has closed.
```
amb26: {Promise}

amb26:
I tried this out with the current Windows 10 vagrant box and found that neither this branch or master are capable of starting the on-screen keyboard when logged on as a normal user by logging in as alice - I get the dialog saying "Could not start On-Screen Keyboard" - does this require a further fix? Both this branch and master work fine when logged on as the standard vagrant admin user.

amb26:
Also, as per @stegru's comment above, I can't reproduce this issue by using elod on the Windows 10 vagrant box, since it appears that variety of the issue is Windows 8.1 only. Is there any variant of this issue I can reproduce on the vagrant box?

stegru:
Did you also also apply the PR for universal? It has an update to the solutions registry.

https://github.com/GPII/universal/pull/479

stegru:
And to reproduce it, you need to enable UAC, or perhaps create a new user who isn't an admin and log in as that.

gpii/node_modules/WindowsUtilities/WindowsUtilities.js
```diff
@@ -654,5 +664,86 @@ windows.enumerateWindows = function (callback) {
...
+            conditionMet = (returnValue === options.conditionValue);
+        }
+
+        if (conditionMet) {
+            promise.resolve();
+        } else {
+            // Check the time.
+            var timedout = false;
+            if (typeof(options.timeout) === "number") {
+                var hr = process.hrtime(startTime);
```
amb26: If there ends up being further work on this branch, do you think you could rewrite this util so that it uses Date.now() rather than the node platform API? It would be cool for it to be a drop-in to get into Infusion someday, since I don't see why it couldn't be universally portable. If there are no other problems with this branch, this is no biggie and this work can wait till another time

stegru: Yes. I just used that method to defend against clock changes.

stegru:
I'm having a hard time reproducing "Could not start On-Screen Keyboard" on either this branch, or master. (master already has a fix, and this branch also fixes it, but differently), latest VM.

Because it also fails on master, it implies this PR may not be the problem. 

A few things you can try:

- As the normal user, run osk.exe - eliminate the possibility of it just being broke.
- Double check the entry in win32.json for osk. This branch should call enableRegisteredAT, and master should execute explorer (with osk.exe as a parameter). Also, change it to notepad.exe, as a sanity check.

stegru:
So, 10 minutes of getting my previous comment off my chest, I can now reproduce it.

It seems to be a problem with Windows itself (it occurs without GPII)

stegru:
This is a Windows bug, where OSK does not start for a non-admin user when UAC is enabled. https://issues.gpii.net/browse/GPII-2142

PR #101, opened 2016-11-18
GPII-2148 Removing unused debugging blocks.

sgithens:
https://issues.gpii.net/browse/GPII-2148


sgithens:
@amb26 Do you think this would be quick to merge?  These blocks look like clear debugging output that someone left in the code, so I'm hoping it's non-controversial to remove them.

PR #102, opened 2016-12-13
GPII-1716: Fontsize support for Windows 10

stegru:
Only implemented for Windows 10 - it works on lower versions in the sense that it does nothing (it doesn't crash).

"gert" is a good person to use, and it requires a resolution of at least 1024x768.


javihernandez:
ok to test

gtirloni:
ok to test

gtirloni:
ok to test

stegru:
The acceptance tests for this are going to fail, because it relies on https://github.com/GPII/universal/pull/493.

https://github.com/GPII/windows/pull/102/commits/65536ad7beefe5e422df61dc1bc41e3aa86b9704 shows it passing, as I (temporarily) removed the DPI setting block from the settings handler output.

stegru:
Closed for https://issues.gpii.net/browse/GPII-2226

PR #103, opened 2016-12-18
GPII-2085 - Provision Node.js LTS branch by default

gtirloni:
Traditionally, the Node.js package published in Chocolatey tracked the latest unstable version. To ensure we were using the LTS versions, the version had to be tracked manually.

On Dec 17th, a new package was published "nodejs-lts", which tracks the LTS versions automatically. This is equivalent to the "nodejs.install" package.

The downside is that this package started on version 6.9.2 so older versions are only available in the "nodejs.install" package. This should not be a problem for future work but might impact developers still using older versions.


gtirloni:
ok to test

gtirloni:
ok to test

kaspermarkus:
We will not merge this because of issues like the npm@5 where node_modules named folders always are removed.

PR #104, opened 2016-12-19
GPII-2073: Persona Settings return on windows lock screen (again)

stegru:
Small change to stop the "deactivate AT" routine from being delayed.


colinbdclark:
ok to test

PR #105, opened 2017-01-09
GPII-2195 : Reverting GPII-1996 (8b02552) because it's not needed anymore.

gloob:
It's fixed in upstream and we are using fixed versions now so it's not needed we can drop it.


javihernandez:
Thanks for keeping track of this issue @gloob ;)

PR #106, opened 2017-01-18
GPII-2200: Debug logging (for HST)

stegru:
Added logging for the following:
- Some build/version identifier
- Is the user an admin?
- If the 64-bit node could be used, then the bitness of the process.
- Running processes
- Dump of the environment variables

It always logs this information, so it might be better to have a flag somewhere (or perhaps make use of the current log level setting)

### DIFF
--- a/gpii/node_modules/debugLogger/debugLogger.js
+++ b/gpii/node_modules/debugLogger/debugLogger.js
@@ -194,7 +194,7 @@ gpii.windows.debugLog.logRunningProcesses = function (options) {
 
     var output = outputLines.join("\n");
 
-    gpii.windows.debugLog.logMessage(output, options);
+    gpii.windows.debugLog.logMessage("Running Processes:\n" + output, options);
 
     return output;
 };
--- a/gpii/node_modules/debugLogger/debugLogger.js
+++ b/gpii/node_modules/debugLogger/debugLogger.js
@@ -19,7 +19,10 @@
 "use strict";
 
 var fluid = require("universal");
+var ref = require("ref");
+
 require("../WindowsUtilities/WindowsUtilities.js");
+require("../processHandling/processHandling.js");
 require("../registrySettingsHandler");
 
 var gpii = fluid.registerNamespace("gpii");
@@ -28,7 +31,7 @@ var windows = fluid.registerNamespace("gpii.windows");
 var debugLogLevel = fluid.logLevel.IMPORTANT;
 
 gpii.windows.logDebugInfo = function () {
-    var funcs = [ "logEnvironment", "logAdministrator", "logArchitecture"];
+    var funcs = [ "logEnvironment", "logAdministrator", "logArchitecture", "logRunningProcesses"];
 
     for (var n = 0, len = funcs.length; n < len; n++) {
         gpii.windows[funcs[n]]();
@@ -98,5 +101,44 @@ gpii.windows.logArchitecture = function (options) {
     return output;
 };
 
+/**
+ * Logs the currently running processes.
+ *
+ * @param [options] Options
+ * @param {String} options.quiet Set to true to create no output.
+ * @return {String} Returns a string containing the debug information.
+ */
+gpii.windows.logRunningProcesses = function (options) {
+    options = options || {};
+
+    var procs = windows.getRunningProcesses();
+
+    var outputLines = fluid.transform(procs, function (proc) {
+        // Get the full path of the executable
+        var exePath = proc.exeFile;
+        var hProcess = windows.kernel32.OpenProcess(windows.API_constants.PROCESS_QUERY_LIMITED_INFORMATION, 0, proc.pid);
+        if (hProcess) {
+            var size = ref.alloc(windows.types.DWORD);
+            size.writeUInt32LE(0x800, 0);
+            var path = new Buffer(0x800);
+            path.fill(0);
+            var success = windows.kernel32.QueryFullProcessImageNameW(hProcess, 0, path, size);
+            if (success) {
+                exePath = windows.fromWideChar(path);
+            }
+            windows.kernel32.CloseHandle(hProcess);
+        }
+
+        return exePath;
+    });
+
+    var output = outputLines.join("\n");
+
+    if (!options.quiet) {
+        fluid.log(debugLogLevel, output);
+    }
+
+    return output;
+};
 
 gpii.windows.logDebugInfo();
--- a/gpii/node_modules/debugLogger/debugLogger.js
+++ b/gpii/node_modules/debugLogger/debugLogger.js
@@ -8,9 +8,11 @@
  * Licensed under the New BSD license. You may not use this file except in
  * compliance with this License.
  *
- * The research leading to these results has received funding from the European Union's
- * Seventh Framework Programme (FP7/2007-2013)
- * under grant agreement no. 289016.
+ * The R&D leading to these results received funding from the
+ * Department of Education - Grant H421A150005 (GPII-APCP). However,
+ * these results do not necessarily represent the policy of the
+ * Department of Education, and you should not assume endorsement by the
+ * Federal Government.
  *
  * You may obtain a copy of the License at
  * https://github.com/GPII/universal/blob/master/LICENSE.txt
--- a/gpii/node_modules/debugLogger/debugLogger.js
+++ b/gpii/node_modules/debugLogger/debugLogger.js
@@ -218,5 +218,3 @@ gpii.windows.debugLog.logArchitecture = function (options) {
 
     return output;
 };
-
-gpii.windows.debugLog.logDebugInfo();
--- a/gpii/node_modules/debugLogger/debugLogger.js
+++ b/gpii/node_modules/debugLogger/debugLogger.js
@@ -20,6 +20,7 @@
 
 var fluid = require("universal");
 var ref = require("ref");
+var fs = require("fs");
 
 require("../WindowsUtilities/WindowsUtilities.js");
 require("../processHandling/processHandling.js");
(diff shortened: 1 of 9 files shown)


gpii/node_modules/debugLogger/debugLogger.js
```diff
@@ -0,0 +1,208 @@
...
+var fs = require("fs");
+
+require("../WindowsUtilities/WindowsUtilities.js");
+require("../processHandling/processHandling.js");
+require("../registrySettingsHandler");
+
+var gpii = fluid.registerNamespace("gpii");
+var windows = fluid.registerNamespace("gpii.windows");
+
+var debugLogLevel = fluid.logLevel.IMPORTANT;
```
amb26: Stick this in a public namespaced constant so that there is a chance to configure it - e.g. the whole contents of this file can go into gpii.windows.debugLog

gpii/node_modules/debugLogger/debugLogger.js
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    return output;
+};
+
+/**
+ * Logs the currently running processes.
+ *
+ * @param [options] Options
+ * @param {String} options.quiet Set to true to create no output.
```
amb26: Accept a user override for the debugLog value in this structure as well - this can replace the "quiet" value

gpii/node_modules/debugLogger/debugLogger.js
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    return output;
+};
+
+/**
+ * Logs the version of GPII
+ *
+ * @param [options] Options
+ * @param {String} options.quiet Set to true to create no output.
```
amb26: Ditto

gpii/node_modules/debugLogger/debugLogger.js
stegru: done

gpii/node_modules/debugLogger/debugLogger.js
stegru: done

amb26:
ok to test

gpii/node_modules/debugLogger/debugLogger.js
```diff
@@ -0,0 +1,244 @@
...
+ */
+gpii.windows.debugLog.logEnvironment = function (options) {
+    options = options || {};
+
+    var output = "";
+    fluid.each(process.env, function (value, name) {
+        output += name + "=" + value + "\n";
+    });
+
+    if (!options.quiet) {
```
amb26: We can get rid of all the options.quiets - which in any case aren't documented

gpii/node_modules/debugLogger/debugLogger.js
```diff
@@ -0,0 +1,244 @@
...
+        windows.debugLog.logMessage(output, options);
+    }
+
+    return output;
+};
+
+/**
+ * Logs the CPU architecture the process is running on.
+ *
+ * @param [options] Options
```
amb26: Not that we have anything that parses these, but our "house style" for these comments is slightly different (param name before type, etc.) and looks like this: https://github.com/GPII/universal/blob/master/gpii/node_modules/gpii-oauth2/gpii-oauth2-datastore/src/DataSource-pouchDB.js#L70

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -60,6 +60,24 @@ gpii.windows.killProcessByName = function (filename) {
...
+        if (all) {
+            togo = fluid.transform(processes, function (proc) {
+                return proc.pid;
+            });
+        } else {
+            togo = processes[0].pid;
+        }
+    }
+
+    return togo;
```
amb26: Looks like the original implementation of this function was left behind as unreachable code after this return statement.

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -103,6 +121,44 @@ gpii.windows.findProcessByName = function (filename, all) {
     return all ? matches : null;
 };
 
+gpii.windows.getRunningProcesses = function () {
```
amb26: This factored-out function now needs its own doc comment

package.json
```diff
@@ -23,5 +23,8 @@
     "keywords": ["gpii", "accessibility", "settings", "fluid", "IoC", "Inversion of Control", "configuration", "evented"],
     "repository": "git://github.com/GPII/windows.git",
     "main": "./gpii.js",
-    "engines": { "node" : ">=6.3.1" }
+    "engines": { "node" : ">=6.3.1" },
+    "scripts": {
+        "preinstall": "git describe --long --always --all --dirty > .version"
```
amb26: This is like one of those famous BLACK PERL poems ....
https://en.wikipedia.org/wiki/Black_Perl

gpii/node_modules/debugLogger/debugLogger.js
```diff
@@ -0,0 +1,244 @@
...
+var fluid = require("universal");
+var ref = require("ref");
+var fs = require("fs");
+
+require("../WindowsUtilities/WindowsUtilities.js");
+require("../processHandling/processHandling.js");
+require("../registrySettingsHandler");
+
+var gpii = fluid.registerNamespace("gpii");
+var windows = fluid.registerNamespace("gpii.windows");
```
amb26: For consistency, you can axe this variable and the next assignment - since almost everything in this file refers to gpii.windows.Xxxxx
So this can just read
var gpii = fluid.registerNamespace("gpii");
fluid.registerNamespace("gpii.windows.debugLog");

gpii/node_modules/debugLogger/debugLogger.js
stegru: done

gpii/node_modules/debugLogger/debugLogger.js
stegru: done

gpii/node_modules/processHandling/processHandling.js
stegru: done

gpii/node_modules/processHandling/processHandling.js
stegru: done

package.json
stegru: haha!

gpii/node_modules/debugLogger/debugLogger.js
stegru: done.

package.json
amb26: It's a shame, since this was a great idea - but when run in our vagrant box, this script explodes with the following failure:

fatal: Invalid symlink 'C:/vagrant': Function not implemented

One (somewhat insane) possibility for extracting this information is:
i) inspecting any file named .git\HEAD , if it exists
ii) if it happens to take the form ref: xxxxxxx then indirecting into the file .git/xxxxxxx and taking the commit hash from there?
If we like this idea, the functionality could be packaged in a node script named, say, extractGitRevision.js (together with a comment explaining why it is necessary).
Then we've got the problems of how/when to invoke it. This project might end up aggregated in several ways. I imagine for the HST, this module will be consumed from npm as a dependency of gpii-app, e.g. https://github.com/michelled/gpii-app/blob/GPII-1596/package.json#L27 in which case (even if it is a git dependency, I believe), the .git directory will be absent.

So whilst this was a reasonable idea, I think the best plan is simply to axe this idea of extracting the git revision from the filesystem, and instead just depend that the person who published this made a usefully identifying version in the NPM package. This is easy enough to do if we use the https://github.com/fluid-project/fluid-publish project to source gpii-windows from NPM - which we should make a note to do for the HST release.

gpii/node_modules/debugLogger/debugLogger.js
```diff
@@ -0,0 +1,230 @@
...
+            }
+            gpii.windows.kernel32.CloseHandle(hProcess);
+        }
+
+        return exePath;
+    });
+
+    var output = outputLines.join("\n");
+
+    gpii.windows.debugLog.logMessage(output, options);
```
amb26: Should be a header here to indicate that "Running Processes:" are being logged

package.json
stegru: Inspecting things in .git doesn't feel right... but yes, .git isn't there anyway when it's a dependency (as I was looking at doing this in the installer script).

fluid-publish gives enough info, I beleive.

gpii/node_modules/debugLogger/debugLogger.js
stegru: done

amb26:
In itself this looks ok to me, but amazingly it seems to harbour some problem which causes the acceptance tests to fail, through being unable to kill the magnifier. I can reproduce this locally too, as well as it appearing in CI. It's hard to imagine anything you have done via the extra Windows API calls has triggered this, but it really seems to be the case! Shucks.

amb26:
ok to test

stegru:
Even though the processHandling code wasn't being used in the test, it was being used elsewhere (somehow).

amb26:
This is officially "okie dokes" and I've merged it into https://github.com/GPII/windows/tree/hst-2017 which will be our HST branch - it can probably go into trunk as well but we should just chat this over at arch on Wednesday

gpii/node_modules/debugLogger/debugLogger.js
```diff
@@ -0,0 +1,220 @@
...
+ * - Version
+ * - Environment variables
+ * - Whether or not the process is running as Administrator
+ * - The CPU architecture
+ * - Running processes
+ *
+ * @param options {Object} [Optional] Options
+ * @param options.logLevel {String} The log level to use, instead of windows.debugLog.logLevel (See fluid.logLevelsSpec)
+ */
+gpii.windows.debugLog.logDebugInfo = function (options) {
```
amb26: I just spotted that this main driver function has no test case

stegru:
This is in HST.

PR #107, opened 2017-01-25
GPII-2091: Provisioning scripts and tests now runnable from gpii-app

kaspermarkus:
Fixed the provisioning scripts so they're no longer dependent on being in a specific location in the VM. This means they now can be run from the gpii-app. Also changed the tests to not depend on a node.exe process to be running - since this wont be the case when using atom electron.


michelled:
@gloob @amb26 can you take a look at this? The gpii-app is depending on the fixes in this pull request.

michelled:
@kaspermarkus are the tests all passing for you? I'm getting a failure in the processHandling tests.

provisioning/Build.ps1
```diff
@@ -1,18 +1,26 @@
 <#
   This script execute all the needed steps to build all the Windows GPII
   components.
+
+  If the -skipNpm flag is provided, no npm install will be run
+  If the script is copied and run from a temporary folder (like when running via vagrant)
+     the -originalBuildScriptPath parameter should be passed with the path to the original
```
amb26: Can we explain why? What use is made of the original folder. This material looks like it is useful to people using the scripts rather than just developers, and so should go into a README with other similar information.

gpii/node_modules/processHandling/test/testProcessHandling.js
```diff
@@ -51,7 +52,7 @@ jqUnit.test("Testing findProcessByName", function () {
     jqUnit.assertEquals("Process should not have been found running", null, pid);
 
     // Find a process that is running.
-    pid = gpii.windows.findProcessByName("node.exe");
+    pid = gpii.windows.findProcessByName(alwaysExistingProcess);
```
amb26: Why was this changed? The change is probably fine, but I'm interested in what the situation was in which detection of node.exe wasn't sufficient.

Vagrantfile
```diff
@@ -32,17 +32,12 @@ Vagrant.configure(2) do |config|
...
-    $destinationDir = Join-Path $env:SystemDrive "tmp"
-    cp $moduleLocation $destinationDir
   SHELL
-  
-  config.vm.provision "shell", path: "provisioning/Chocolatey.ps1"
-  config.vm.provision "shell", path: "provisioning/Npm.ps1"
-  config.vm.provision "shell", path: "provisioning/Build.ps1"
-  config.vm.provision "shell", path: "provisioning/Installer.ps1"
+
+  # Provide original script path for use in relative paths, since vagrant copies the script to a temporary location
```
amb26: Why does vagrant copy them to a temporary location?

kaspermarkus:
ping @gloob

kaspermarkus:
This is now outdated and have been improved significantly in other branches by the great Spaniards

PR #108, opened 2017-02-08
GPII-2235 & GPII-2091 : Updating reference to universal where singleInstance is present and updated status of provisioning scripts.

gloob:
GPII-2235 & GPII-2091 : Updating reference to universal where singleInstance is present and updated status of provisioning scripts.


javihernandez:
merged into hst-2017

PR #109, opened 2017-02-08
GPII-2277 : Executing nodevars.bat correctly in PS1 environment.

gloob:
Now we have available all the nodejs and npm environment vars in the first execution of the VM.


javihernandez:
Thanks @gloob!
Merged into both master and hst-2017 branches

PR #112, opened 2017-02-21
GPII-2125: Use the closest available resolution when setting it

javihernandez:
As described in GPII-2125, since we can't set the screen resolution to something that is not recognized as valid (VirtualBox does its own black magic when resizing), we at least should try to restore the resolution to the closest valid one.

### DIFF
--- a/gpii/node_modules/displaySettingsHandler/src/displaySettingsHandler.js
+++ b/gpii/node_modules/displaySettingsHandler/src/displaySettingsHandler.js
@@ -108,42 +108,56 @@ windows.display.getScreenResolution = function () {
 };
 
 /*
+ * Gets available resolutions
+ *
+ * @return {list} of resolutions
+ */
+windows.display.getAvailableResolutions = function () {
+    var index = 0;
+    var dm = windows.display.createDevModeStruct();
+    var availableResolutions = [];
+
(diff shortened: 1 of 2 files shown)


gpii/node_modules/displaySettingsHandler/src/displaySettingsHandler.js
```diff
@@ -108,42 +108,56 @@ windows.display.getScreenResolution = function () {
...
-                        // the user.
-                        //
-                        return true;
-                    } else {
-                        // Return codes explained at https://msdn.microsoft.com/en-us/library/dd183411.aspx
-                        fluid.fail("Couldn't set screen resolution, return code was " + lRet);
-                    }
-                }
+        var dmCurrent = windows.display.createDevModeStruct();
+        var closestResolution = windows.display.findClosestResolution(newRes);
```
amb26: I think we had decided that we would do this calculation on "get" rather than on "set" - because otherwise the LifecycleManager/snapshotter could become confused as a result of their apparent inability to set the settings. It should be the case that after we do a "set", the results of "get" should agree with what we have just set. So we should have "get" report the nearest available resolution, and "set" to only allow the system to be set to one of the valid ones.

javihernandez: Right, I've applied this change and added back the test that ensures that we can restore to the (closest) original resolution.

javihernandez:
ok to test

PR #113, opened 2017-02-27
GPII-2302: Replace calls to shelljs.exec

javihernandez:
This problem only happens when running the tests from within electron, more info at https://issues.gpii.net/browse/GPII-2302


javihernandez:
FYI, we've had a discussion about this on IRC (https://botbot.me/freenode/fluid-work/2017-02-27/?msg=81632399&page=1).

javihernandez:
I've just replaced all references to shelljs.exec with child_process calls.
@amb26 and @stegru, are we happy with this?

PR #114, opened 2017-03-06
GPII-2315: Magnifier doesn't apply subsequent settings

stegru:
Only the settings of the first instance of magnifier are applied. Subsequent instances continue to use those set first.

This fix performs some clean-up that magnifier would have done if it was closed nicely.

### DIFF
--- a/gpii/node_modules/registeredAT/registeredAT.js
+++ b/gpii/node_modules/registeredAT/registeredAT.js
@@ -54,6 +54,13 @@ gpii.windows.registeredAT = fluid.freezeRecursive({
     registeredATKey: {
         baseKey: "HKEY_LOCAL_MACHINE",
         path: "SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion\\Accessibility\\ATs"
+    },
+    /**
+     * The registry key where the settings to be transferred to the secure desktop are put.
+     */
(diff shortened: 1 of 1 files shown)


stegru:
ok to test

gpii/node_modules/registeredAT/registeredAT.js
```diff
@@ -208,18 +215,29 @@ gpii.windows.startRegisteredAT = function (name, options) {
...
         exeName = atInfo.ATExe;
     }
 
-    return gpii.windows.killProcessByName(exeName);
+    var promiseTogo = gpii.windows.killProcessByName(exeName);
+
+    // Remove the registry key that's used to transfer settings to the secure desktop, otherwise these settings
+    // will override any future settings. Really, the AT should do this when it closes, unfortunately it doesn't get
+    // closed cleanly so it is getting done here instead.GPII-2315.
+    if (atInfo.CopySettingsToLockedDesktop) {
```
amb26: Any signs of where, if anywhere, this flag is documented?

stegru: Yes, in https://msdn.microsoft.com/library/windows/desktop/bb879984.aspx, which is mentioned in the comment near the start of the file.

PR #117, opened 2017-03-28
GPII-2338: Unable to log out if magnifier is closed manually

stegru:


javihernandez:
@stegru I think you meant GPII-2339. Can you re-open this according to the right JIRA number?
Also, maybe we should provide a test case to cover this. Other than that, I'm fine with it. @amb26, any thoughts on this?

PR #118, opened 2017-03-30
GPII-2339: Unable to log out if magnifier is closed manually

stegru:


javihernandez:
Didn't see that the pull request was against hst-2017 and I'd say this should land master too. @stegru, can you submit a pull request of the same branch but this time against master?

stegru:
Done: #119

PR #119, opened 2017-03-31
GPII-2339: Unable to log out if magnifier is closed manually

stegru:


gtirloni:
ok to test

kaspermarkus:
I've reviewed the code, and that all looks good. I am unable to reproduce GPII-2339 on my machine, so @javihernandez has agreed to check that this fixes the issue. Once done either @javihernandez or myself will merge

stegru:
ok to test

PR #128, opened 2017-07-13
GPII-2483: Updated reference to universal

javihernandez:
I'm also removing the nan dependency since it is no longer needed after node 6.8.0 was released.


amb26:
Closed for #129

PR #139, opened 2017-09-01
Update with last version of master and workaround for GPII-2084

javihernandez:
Update with last version of master and workaround to work with electron-edge when on electron.


javihernandez:
Closed for https://github.com/GPII/windows/pull/143

PR #144, opened 2017-09-07
GPII-2582: Fix failure in gpii-app tests

stegru:
This prevents the metrics capture starting while testing.


stegru:
ok to test

stegru:
ok to test

PR #146, opened 2017-09-12
Fix processes bridge to run the same in PowerShell as in Command Prompt

klown:
@javihernandez, here's a pull request to fix the unit test in gpii-app. 

The problem: when run under PowerShell, the C# code returns slightly different information compared to when run under the older command prompt.  CI for the windows repo uses the command prompt for the unit tests (and they all pass).  Gpii-app, however, is launched via a PowerShell script, and the difference causes that one unit test to fail.  I've normalized the output of the C# code such that it should no longer matter.

I hope this does the trick.

### DIFF
--- a/gpii/node_modules/processReporter/dotNetProcesses.csx
+++ b/gpii/node_modules/processReporter/dotNetProcesses.csx
@@ -84,6 +84,14 @@ public class Startup
         propValue = process.GetPropertyValue("CommandLine");
         if (propValue != null) {
             procInfo.argv = makeArgv(propValue.ToString());
+
+            // In some cases the result is just the command, e.g., "Notepad".
+            // In others, it is the full path and may include the ".exe"
+            // extension.  Normalize it to always be the full path, if there is
+            // one.
+            if (procInfo.fullPath.Length > 0) {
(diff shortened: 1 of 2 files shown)


klown:
ok to test

gtirloni:
ok to test

klown:
ok to test

klown:
@javihernandez The changes above fix one of the test failures.  But, there is still one failure, and it's tricky to diagnose.  It's not a test failure, but something to do when no argument is passed to `getProcessList()`.  It works fine in the windows branch.  Still diagnosing in gpii-app.

klown:
Update @javihernandez: I have figured out the other bug, but haven't finished the solution.  I'll take it up again in the morning.

klown:
ok to test

javihernandez:
ok to test

klown:
ok to test

klown:
ok to test

javihernandez:
This won't pass until we update the vagrant box OR get [this](https://github.com/GPII/windows/pull/148) merged into master

javihernandez:
@klown can you update this pull request with master?

PR #150, opened 2017-09-19
GPII-2583: Metrics improvements

stegru:
Improvements:
- Mouse activity.
- Inactivity detection.
- Solution IDs that are applied.
- Keystroke times logged better
- Only capturing metrics when keyed in
- Error reporting contains the stack
- Unit tests are a lot better

### DIFF
--- a/gpii/node_modules/windowsMetrics/src/windowsMetrics.js
+++ b/gpii/node_modules/windowsMetrics/src/windowsMetrics.js
@@ -87,12 +87,12 @@ fluid.defaults("gpii.windowsMetrics", {
                 precision: 5000
             },
             keyboard: {
-                // Number of records per log line (falsey for no limit).
-                maxRecords: 500,
                 // Minimum typing session time, in milliseconds.
                 minSession: 30000,
                 // The time a session will last with no activity, in milliseconds.
-                sessionTimeout: 60000
+                sessionTimeout: 60000,
+                // Minimum number of keys in a typing session time.
+                minSessionKeys: 30000
             }
         },
         state: {
@@ -101,7 +101,12 @@ fluid.defaults("gpii.windowsMetrics", {
                 keyboardHookHandle: null,
                 lastKeyLogged: null,
                 lastKeyTime: 0,
-                times: []
+                // Timestamp of typing session start.
+                sessionStart: null,
+                // Number of keys in the typing session.
+                keyCount: 0,
+                // Numnber of special keys
+                specialCount: 0
             }
         }
     }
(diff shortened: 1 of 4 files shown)


stegru:
ok to test

gpii/node_modules/windowsMetrics/src/windowsMetrics.js
```diff
@@ -287,31 +282,51 @@ windows.checkActiveWindow = function (that) {
...
+            windows.user32.SetWindowsHookExW(WH_KEYBOARD_LL, windows.keyboardHookCallback, hModule, 0);
+
+        if (!that.model.state.input.keyboardHookHandle) {
+            var errCode = windows.kernel32.GetLastError();
+            windows.stopInputMetrics(that);
+            fluid.fail("SetWindowsHookExW did not work (keyboard). win32 error: " + errCode);
+        }
+
+        // Start the mouse hook
+        that.model.state.input.mouseHookHandle =
```
amb26: Must not write material directly into a component's model. If there is random mutable stuff that you don't need a ChangeApplier facility for, you can just stick it on a top-level component member. Otherwise, if this really is suitable model material (that is, you expect it to serialise to and from JSON correctly) you can say that.applier.change("state.input.mouseHookHandle", etc. )

gpii/node_modules/windowsMetrics/src/windowsMetrics.js
```diff
@@ -321,17 +336,23 @@ windows.startKeyboardMetrics = function (that) {
...
-    that.model.state.keyboard.keyboardHookHandle = null;
-    that.model.state.keyboard.keyboardHookCallback = null;
-    // flush the log.
-    if (that.model.state.keyboard.times) {
-        windows.logKeyTimings(that);
-        that.model.state.keyboard.times = [];
+    if (state.mouseHookHandle) {
+        windows.user32.UnhookWindowsHookEx(state.mouseHookHandle);
+    }
+    state.keyboardHookHandle = null;
```
amb26: Similarly - if this stuff is not intended to be modelised, just stick it on an ordinary member named "state" - else use the changeApplier to manipulate it.

gpii/node_modules/windowsMetrics/test/WindowsMetricsTests.js
```diff
@@ -827,46 +817,56 @@ jqUnit.asyncTest("Testing keyboard metrics: keyboardHook", function () {
...
+    var logFile = gpii.tests.metrics.setLogFile();
+
+    var windowsMetrics = gpii.tests.metrics.windowsMetricsWrapper({
+        members: {
+            logFilePath: logFile
+        }
+    });
+
+    // Make the inactivity timeout very quick
+    windowsMetrics.model.config.input.inactiveTime = 1;
```
amb26: As above - must not randomly bash on a component's model, future revisions of the framework will probably prevent it anyway

gpii/node_modules/windowsMetrics/src/windowsMetrics.js
stegru: No reason for it to be a model.

gpii/node_modules/windowsMetrics/src/windowsMetrics.js
stegru: done.

gpii/node_modules/windowsMetrics/test/WindowsMetricsTests.js
stegru: yep.

PR #120, opened 2017-04-20
GPII-1230, GPII-1235, GPII-442: Running application detection and support for more advanced MM

kaspermarkus:
This branch is compatible with the advanced MM branch GPII-1230 of universal (https://github.com/GPII/universal/pull/507). It contains the (relevant) code from Josephs GPII-442 branch

PR #152, opened 2017-09-22
GPII-2521: Merging of Morpher and Simplification into upstream

JavierJF:
This pull is blocked by [this other pull request.](https://github.com/GPII/universal/pull/554)


amb26:
Cheers, @JavierJF - I've updated the linting configuration in master to match that in our other projects. Please could you regenerate this pull against master since it has picked up some changes from the process reporter work, and then make sure it passes linting - thanks!

gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
```diff
@@ -260,6 +315,96 @@ gpii.windows.spi.populateResults = function (payload, isNewValue, isGetting, res
...
+        return s.constructor === String;
+    }
+    return typeof s === "string";
+}
+
+/**
+ * @param {*} s String to count its words.
+ * @return The number of words in the string parameter.
+ */
+function countWords(s) {
```
amb26: Interesting algorithm. Given it handles so many cases, we should provide tests for them - or else simplify it so that it handles a more restricted range of possibilities. However, I'm concerned by the asymmetry between "countWord" and "splitInWord". Really, they should operate the same algorithm or at least algorithms which are complementary.
Whatever we do in this area should also be exported as a public function (which you would need to do anyway in order to write tests - such wonders of test-oriented development!)

gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
```diff
@@ -260,6 +315,96 @@ gpii.windows.spi.populateResults = function (payload, isNewValue, isGetting, res
     }
 };
 
+/**
+ * @param {String} s Returns true if value is a string, false otherwhise.
+ */
+function isString(s) {
```
amb26: I am baffled that this function is necessary - what are these strings which fail the typeof check and yet have a constructor which is a string?

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -296,6 +297,120 @@ gpii.windows.closeProcessByName = function (filename, options) {
...
+ * @param processToWait The process that is going to be closed when the command executes.
+ * @param processToClose The process for closing after the previous one has exited.
+ * @param options.timeout How long to wait for the process to die, in milliseconds.
+ * @param options.cleanOnly true to reject if the process can't be closed cleanly; false to force the termination.
+ * @param options.exitCode {number} The exit code the application should return.
+ * @param options.restart Restart the processToClose instead of just closing it.
+ * @return {promise}
+ */
+gpii.windows.closeAfterExitCommand = function (command, args, processToWait, processToClose, options) {
+    var defaultOptions = {
```
amb26: Factor this repeated definition into a public one, perhaps even a non-component defaults block

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -296,6 +297,120 @@ gpii.windows.closeProcessByName = function (filename, options) {
     return promiseTogo;
 };
 
+/**
+ * @param command The command to be executed for closing an application.
```
amb26: These two functions are nearly identical and should be factored together. Also, there should be a clear explanation of the purpose and workflow of the function.

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -296,6 +297,120 @@ gpii.windows.closeProcessByName = function (filename, options) {
...
+
+    options = fluid.extend(defaultOptions, options);
+
+    var promise = fluid.promise();
+
+    var closeCommand = gpii.launch.spawn(command, args);
+
+    // Linter (Bug) Workaround: Linter fails with arrow style lambda with multiple parameters.
+    closeCommand.on('exit', function (code, signal) {
+        // Linter (Bug) Workaround: Linter fails with arrow style lambda without parameters.
```
amb26: It is not a bug, it is simply that such ES6 neologisms are not part of our project standard :)
I think you will find that when you factor these functions better, the desirability of them will fade away a bit ...

amb26:
I think we will have to leave the wallpapers out of the repo since they don't meet our licence conditions. Let's just go with solid colour backgrounds, or open licenced placeholders, for now, and if this is queried we can start the wider licencing discussion.

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -296,6 +297,120 @@ gpii.windows.closeProcessByName = function (filename, options) {
...
+            gpii.windows.closeProcessByName(processToClose, {cleanOnly: false, timeout: 150})
+                .then(function () {
+                    if (options.restart) {
+                        var b64Cmd = process.env.windir + "\\sysnative\\cmd.exe";
+                        child_process.execFile(b64Cmd, ["/k", processToClose]);
+                        promise.resolve(true);
+                    } else {
+                        promise.resolve(true);
+                    }
+                }, (err) => {
```
amb26: This ES6 neologism also fails linting (not a bug : P )

amb26:
For me the Unit tests for this branch fail with the following: 

07:37:32.967:  {
    "module": "GPII",
    "event": "Error.Fail",
    "level": "FAIL",
    "data": {
        "error": {
            "message": "gpii.windows.spi.applySettings: Invalid PvParam payload '\u0000\u0000'",
            "stack": "Error\n    at gpii.eventLog.logError (\\\\vboxsrv\\vagrant\\node_modules\\universal\\gpii\\node_modules\\eventLog\\src\\eventLog.js:187:28)\n    at invokeInvoker [as logError] (\\\\vboxsrv\\vagrant\\node_modules\\infusion\\src\\framework\\core\\js\\FluidIoC.js:1736:29)\n    at Function.gpii.eventLog.gotError (\\\\vboxsrv\\vagrant\\node_modules\\universal\\gpii\\node_modules\\eventLog\\src\\eventLog.js:208:21)\n    at \\\\vboxsrv\\vagrant\\node_modules\\universal\\gpii\\node_modules\\eventLog\\src\\eventLog.js:197:19\n    at Object.fire (\\\\vboxsrv\\vagrant\\node_modules\\infusion\\src\\framework\\core\\js\\Fluid.js:1533:40)\n    at Object.fluid.fail (\\\\vboxsrv\\vagrant\\node_modules\\infusion\\src\\framework\\core\\js\\Fluid.js:213:32)\n    at Object.gpii.windows.spi.applySettings (\\\\ .... [output suppressed at 1024 chars - for more output, increase fluid.logObjectRenderChars]
07:37:32.987:  jq: FAIL: Module "gpii.tests.acceptance.windows.builtIn.config tests" Test name "Testing os_win7 using default matchmaker" - Message: Assertion failure (see console.log for expanded message): gpii.windows.spi.applySettings: Invalid PvParam payload '  '
07:37:32.987:  jq: Source:     at jqUnit.nodeFailureHandler (\\vboxsrv\vagrant\node_modules\node-jqunit\lib\jqUnit-node.js:34:15)
    at Object.fire (\\vboxsrv\vagrant\node_modules\infusion\src\framework\core\js\Fluid.js:1533:40)
    at Object.fluid.fail (\\vboxsrv\vagrant\node_modules\infusion\src\framework\core\js\Fluid.js:213:32)
    at Object.gpii.windows.spi.applySettings (\\vboxsrv\vagrant\gpii\node_modules\spiSettingsHandler\src\SpiSettingsHandler.js:245:19)
    at gpii.windows.spi.setImpl (\\vboxsrv\vagrant\gpii\node_modules\spiSettingsHandler\src\SpiSettingsHandler.js:514:22)
    at \\vboxsrv\vagrant\node_modules\universal\gpii\node_modules\settingsHandlers\src\settingsHandlerUtilities.js:150:21
    at Object.<anonymous> (\\vboxsrv\vagrant\node_modules\universal\gpii\node_modules\settingsHandlers\src\settingsHandlerUtilities.js:78:20)
    at transformInternal (\\vboxsrv\vagrant\node_modules\infusion\src\framework\core\js\Fluid.js:469:34)
    at Object.fluid.transform (\\vboxsrv\vagrant\node_modules\infusion\src\framewo .... [output suppressed at 1024 chars - for more output, increase fluid.logObjectRenderChars]

gpii/node_modules/processHandling/processHandling.js
JavierJF: Ups... My bad, I will pass again the linter and change the comments about the bug.

gpii/node_modules/processHandling/processHandling.js
JavierJF: Okay, then I will refactor both of then into closeAfterCommand, and change the flow with the payload. What should I do with the 'non-component default block'.

gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
JavierJF: That was just for checking strings that were constructed using 'new operator', then 'typeof check' will fail for them, since it will return 'object'.

gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
JavierJF: Nice! I will start exporting it as a public function, and writing some tests for it. Then we could talk about the cases in which we could reuse it.

gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
amb26: The construction of strings using the new operator is such a crazy practice that there is even an eslint rule to enforce against it https://eslint.org/docs/rules/no-new-wrappers - I notice that this is not in our community rules but I will propose having it added. You may safely assume that noone in our community has done this and simply use typeof (x) === "string"

gpii/node_modules/processHandling/processHandling.js
amb26: You could just say fluid.defaults("gpii.windows.closeAfterCommand", {
   ....
}
and then retrieve them again using the same call. This doesn't offer particularly many benefits other than allowing you to use the same global name for the function as well as for the options.

You can see this pattern used in the framework here: https://github.com/fluid-project/infusion/blob/master/src/framework/core/js/FluidIoC.js#L2305

As a funny historical note, this is all that Infusion defaults ever were, when they were first invented (around 2008)

gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
amb26: Note the following answer: https://stackoverflow.com/questions/5750656/whats-the-point-of-new-stringx-in-javascript - "No native JavaScript object, major library or DOM method that returns a string will return a String object rather than a string value."

JavierJF: Didn't knew about the eslint rule against it, or the global community agreement against it, if it's in our community guidelines I agree also that we could asume that will never happen. Removing it!

amb26:
On my system the increased timeout is still insufficient - 

```
07:03:32.745:  Executing: tasklist /fi "STATUS eq RUNNING" /FI "IMAGENAME eq ClassicShell.exe" | find /I "ClassicShell.exe" /C
07:03:33.603:  Executing: tasklist /fi "STATUS eq RUNNING" /FI "IMAGENAME eq ClassicShell.exe" | find /I "ClassicShell.exe" /C
07:03:34.340:  Executing: tasklist /fi "STATUS eq RUNNING" /FI "IMAGENAME eq ClassicShell.exe" | find /I "ClassicShell.exe" /C
07:03:35.512:  Executing: tasklist /fi "STATUS eq RUNNING" /FI "IMAGENAME eq ClassicShell.exe" | find /I "ClassicShell.exe" /C
07:03:36.729:  Executing: tasklist /fi "STATUS eq RUNNING" /FI "IMAGENAME eq ClassicShell.exe" | find /I "ClassicShell.exe" /C
07:03:37.463:  {
    "module": "metrics",
    "event": "app-active",
    "level": "INFO",
    "data": {
        "exe": "C:\\Windows\\System32\\cmd.exe",
        "duration": 5
    }
}
07:03:37.463:  {
    "module": "metrics",
    "event": "app-launch",
    "level": "INFO",
    "data": {
        "exe": "C:\\Windows\\explorer.exe"
    }
}
07:03:37.666:  Executing: tasklist /fi "STATUS eq RUNNING" /FI "IMAGENAME eq ClassicShell.exe" | find /I "ClassicShell.exe" /C
07:03:37.745:  Test case listener has not responded after 5000ms - at sequence pos 11 of 19 sequence element {
    "event": "{exec}.events.onExecExit",
    "listener": "gpii.test.onExecExit"
} of fixture Testing ClassicShell
07:03:38.338:  Executing: tasklist /fi "STATUS eq RUNNING" /FI "IMAGENAME eq ClassicShell.exe" | find /I "ClassicShell.exe" /C
07:03:38.979:  Executing: tasklist /fi "STATUS eq RUNNING" /FI "IMAGENAME eq ClassicShell.exe" | find /I "ClassicShell.exe" /C
07:03:39.621:  Executing: tasklist /fi "STATUS eq RUNNING" /FI "IMAGENAME eq ClassicShell.exe" | find /I "ClassicShell.exe" /C
07:03:40.247:  Executing: tasklist /fi "STATUS eq RUNNING" /FI "IMAGENAME eq ClassicShell.exe" | find /I "ClassicShell.exe" /C
07:03:40.873:  Executing: tasklist /fi "STATUS eq RUNNING" /FI "IMAGENAME eq ClassicShell.exe" | find /I "ClassicShell.exe" /C
07:03:41.527:  Executing: tasklist /fi "STATUS eq RUNNING" /FI "IMAGENAME eq ClassicShell.exe" | find /I "ClassicShell.exe" /C
07:03:42.166:  Executing: tasklist /fi "STATUS eq RUNNING" /FI "IMAGENAME eq ClassicShell.exe" | find /I "ClassicShell.exe" /C
07:03:42.308:  jq: FAIL: Module "gpii.tests.acceptance.windows.classicShell.config tests" Test name "Testing ClassicShell" - Message: Checking the process with command: [object Object] - at sequence position 11 of 19
07:03:42.308:  jq: Source:     at pok (\\vboxsrv\vagrant\node_modules\infusion\tests\test-core\jqUnit\js\jqUnit.js:112:15)
    at Object.assertTrue (\\vboxsrv\vagrant\node_modules\infusion\tests\test-core\jqUnit\js\jqUnit.js:145:13)
    at gpii.test.onExecExit (\\vboxsrv\vagrant\node_modules\universal\gpii\node_modules\testing\src\Testing.js:182:12)
    at wrapped (\\vboxsrv\vagrant\node_modules\infusion\tests\test-core\utils\js\IoCTestUtils.js:320:26)
    at Object.fire (\\vboxsrv\vagrant\node_modules\infusion\src\framework\core\js\Fluid.js:1533:40)
    at \\vboxsrv\vagrant\node_modules\universal\gpii\node_modules\testing\src\Acceptance.js:55:40
    at ChildProcess.exithandler (child_process.js:213:5)
    at emitTwo (events.js:106:13)
    at ChildProcess.emit (events.js:191:7)
    at maybeClose (internal/child_process.js:877:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:226:5)
```

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -296,6 +297,142 @@ gpii.windows.closeProcessByName = function (filename, options) {
     return promiseTogo;
 };
 
+gpii.windows._restartProcess = function(promise, processToRestart, retries) {
+    var b64Cmd = process.env.windir + "\\sysnative\\cmd.exe";
+    child_process.execFile(b64Cmd, ["/k", processToRestart]);
+
+    gpii.windows.waitForProcessStart(processToRestart, {timeout: 1000, pollDelay:100})
```
amb26: These changes do not pass linting

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -296,6 +297,142 @@ gpii.windows.closeProcessByName = function (filename, options) {
...
+
+    options = fluid.extend(defaultOptions, options);
+
+    var commandRes = gpii.launch.spawn(command, args);
+
+    if (options.commandType === "Open") {
+        fluid.log("Waiting for process start: ");
+        gpii.windows.waitForProcessStart(processToWait, {timeout: 1000, pollDelay: 100})
+            .then(function () {
+                gpii.windows.closeProcessByName(processToClose, {cleanOnly: false, timeout: 150})
```
amb26: These promise handlers are still 100% duplicated

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -296,6 +297,142 @@ gpii.windows.closeProcessByName = function (filename, options) {
     return promiseTogo;
 };
 
+gpii.windows._restartProcess = function(promise, processToRestart, retries) {
```
amb26: The "underscore" convention isn't used within our projects - instead, please supply some doc comments for these functions

amb26:
This failure is in the real system, with what looks like the 15 second timeout. 
Also, we need to remove the copyrighted Microsoft wallpapers and images from this pull before it can be merged.

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -487,6 +350,460 @@ gpii.windows.getProcessPath = function (pid, processHandle) {
...
+        }, options.quitDelay || 3000);
+    } else {
+        promise.resolve();
+    }
+
+    return promise;
+};
+
+/**
+ * Convenience function for trying to close an application by name, with the minimal necessary level of force.
```
stegru: `gpii.windows.closeApplications` already does this - if you pass "high", it will try the low, then medium if low didn't work, then high.

Just passing procPids to closeApplications should be enough.

JavierJF: I realized, but it doesn't have any kind of wait for the process being stopped after any of the different force level calls, so I assumed that it simply 'tries more things' depending on the level of force.

stegru: ok, I understand.

`RmShutdown.async` (line 498) calls back when the process has been stopped, and the "high" method is invoked after a 3 second delay (or options.quitDelay) after "medium".

I think calling this function with "low" is enough, then call `gpii.windows.closeProcessByName` when it resolves.

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -487,6 +350,460 @@ gpii.windows.getProcessPath = function (pid, processHandle) {
...
+                tryCloseWithMediumForce();
+            }
+        );
+
+    return promise;
+};
+
+/**
+ * Close a running application after a command has been executed.
+ *
```
stegru: It took me some time to understand what this actually does. It's not obvious (to me) what it does based on the documentation and signature.

I suggest documenting the primary use of this function (restarts explorer after starting/stopping classic start menu), and renaming it.

My understanding of it is this (ignoring what processA does):
- Starts processA (command)
- Waits for processB (processToWait) to start/stop
- Closes processC (processToClose)

In the only use case of this function, processA and B are both the same executable, so perhaps default `processToWait` to `command` (without the path), and move it into `options`.

This will simplify the function without changing the way it works:

- Starts processA, and wait for it to start/stop.
- Close processC.

Now the function can look like `startAndClose(command, processToClose, options)`, and anything that alters the meaning of the method name is hidden in options.

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -487,6 +350,460 @@ gpii.windows.getProcessPath = function (pid, processHandle) {
...
+        restart: false,
+        retries: 1
+    };
+
+    options = fluid.extend(defaultOptions, options);
+
+    var promise = fluid.promise();
+
+    // Check if the 'commandType' that is going to be executed is valid.
+    var validCommands = ["Open", "Close"];
```
stegru: If there's only ever going to be 2 choices, why not use a boolean and call something like `waitForClose` (or `waitForOpen`)?

(and rather than describing what `command` does, it will say what your code will do)

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -487,6 +350,460 @@ gpii.windows.getProcessPath = function (pid, processHandle) {
...
+    if (validCommands.indexOf(options.commandType) === -1) {
+        promise.reject({
+            isError: true,
+            message: "gpii.windows.closeAfterCommand: Error invalid commandType option '" + options.commandType + "'."
+        });
+        return promise;
+    }
+
+    var keepTrying = function (rPromise, retries) {
+        if (retries <= 0) {
```
stegru: Why does it need to retry? If it failed the first time, why would it succeed the second?

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -487,6 +350,460 @@ gpii.windows.getProcessPath = function (pid, processHandle) {
...
+            if (options.commandType === "Close") {
+                fluid.log("gpii.windows.closeAfterCommand: Timeout for process closing failed after '" + retries + "' retries.");
+                rPromise.reject({
+                    isError: true,
+                    message: "Exceeded the number of retries trying to close process '" + processToClose + "'."
+                });
+            } else if (options.commandType === "Open") {
```
stegru: These 2 blocks are exactly the same, apart from 2 words.

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -487,6 +350,460 @@ gpii.windows.getProcessPath = function (pid, processHandle) {
...
+                                p.reject({
+                                    isError: true,
+                                    message: "gpii.windows.closeAfterCommand: Unable to close application '" + process
+                                });
+                            }
+                        );
+            };
+
+            if (options.commandType === "Open") {
+                gpii.windows.waitForProcessStart(
```
stegru: These two blocks are exactly the same, apart from the function. I suggest doing something like:
```
var waitFor = <start> ? gpii.windows.waitForProcessStart : gpii...Termination;
waitFor(...);
```

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -487,6 +350,460 @@ gpii.windows.getProcessPath = function (pid, processHandle) {
...
+                rPromise.reject({
+                    isError: true,
+                    message: "Error while executing command '" + command + "' with args '" + args + "'.",
+                    error: error
+                });
+            });
+        }
+    };
+
+    keepTrying(promise, options.retries);
```
stegru: It should be ok to use `promise` and (a copy of) `options.retries` in the function.

gpii/node_modules/processHandling/test/testProcessHandling.js
```diff
@@ -303,3 +320,403 @@ jqUnit.asyncTest("Testing getProcessPath", function () {
...
+        } else if (preLevels.length === 2) {
+            jqUnit.assertDeepEq("test: Previous calls should be done in order",
+                preLevels, [1,2]);
+            jqUnit.assertEquals("test: Second call should be done with level 3",
+                levelNum, 3);
+
+            preLevels = [];
+
+            promise.resolve();
+            jqUnit.start();
```
stegru: put this in a `.then` at the end of closeAppWithMinForce

gpii/node_modules/processHandling/test/testProcessHandling.js
```diff
@@ -303,3 +320,403 @@ jqUnit.asyncTest("Testing getProcessPath", function () {
...
+            medium: 2,
+            high: 3
+        };
+
+        var levelNum = levels[level] || levels.low;
+        var preLevels = gpii.tests.windows.mockCloseApplicationLevels;
+
+        var promise = fluid.promise();
+
+        if (preLevels.length === 0) {
```
stegru: Couldn't these go into an array?
```
testData[preLevels.length].levelNum
testData[preLevels.length].preLevels
preLevels.push(preLevels.length+1)
```

They all seem to match length (or +1), so it might be nicer to do that instead.

gpii/node_modules/WindowsUtilities/WindowsUtilities.js
```diff
@@ -724,9 +848,16 @@ windows.accessFlag = function (object, flagName, setValue) {
 windows.arrayToBuffer = function (array, type) {
     var size = ref.coerceType(windows.types[type]).size;
     var buf = new Buffer(array.length * size);
+    var i = 0;
 
-    for (var i = 0; i < array.length; ++i) {
-        ref.set(buf, i * size, array[i], windows.types[type]);
+    if (type === "TCHAR") {
```
stegru: In line 488 of SpiSettingsHandler, this function isn't called if is `TCHAR`.

gpii/node_modules/WindowsUtilities/WindowsUtilities.js
```diff
@@ -742,9 +873,16 @@ windows.arrayToBuffer = function (array, type) {
 windows.bufferToArray = function (buffer, type) {
     var array = [];
     var size = ref.coerceType(windows.types[type]).size;
+    var i = 0;
 
-    for (var i = 0; i < buffer.length / size; ++i) {
-        array[i] = ref.get(buffer, i * size, windows.types[type]);
+    if (type === "TCHAR") {
```
stegru: Like `arrayToBuffer`, this function isn't called if type is `TCHAR`.

See SpiSettingsHandler.js:83

gpii/node_modules/processHandling/processHandling.js
JavierJF: Ok, so you propose to change the implementation to:

 *Call 'closeApplications' with "low", if that isn't enough, call 'closeProcessByName'*

Am I wrong?

gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
```diff
@@ -337,7 +482,11 @@ gpii.windows.spi.getPvParam = function (payload) {
     }
 
     if (payload.options.pvParam.type === "array") {
-        pvParam = gpii.windows.arrayToBuffer(pvParam, payload.options.pvParam.valueType);
+        if (payload.options.pvParam.valueType === "TCHAR") {
```
stegru: An array of TCHARs is technically correct, but I think it might be better to have a new type of "string", and perform the UCS-2/UTF-8 conversion before and after the actual call to SystemParametersInfo

And look at `windows.stringToWideChar` and `windows.stringFromWideChar` for the actual conversion.

gpii/node_modules/processHandling/processHandling.js
JavierJF: I agree it looks more clear with that adjustments, and indeed the target was 'processA' and 'processB' being the same. I will reformat this function with your suggestion.

gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
```diff
@@ -260,6 +315,96 @@ gpii.windows.spi.populateResults = function (payload, isNewValue, isGetting, res
...
+
+            return value;
+        }
+    }
+};
+
+/**
+ * Checks and returns the value for the fWinIni parameter of the SystemParametersInfo function.
+ */
+gpii.windows.spi.getFWinIni = function (payload) {
```
stegru: This isn't called. Was this the duplicate of what

gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
```diff
@@ -181,6 +197,32 @@ gpii.windows.spi.callProblematicSpi = function (pvParamType, action, uiParam, pv
     return promise;
 };
 
+gpii.windows.spi.checkWallpaperPath = function (path) {
```
stegru: `/** */`

gpii/node_modules/processHandling/processHandling.js
JavierJF: Because calling it using a boolean flag like waitForClose seemed to me ambiguos in the false case, that would mean that:

* !waitForClose == waitForOpen *

I didn't considered that equality intuitive.

gpii/node_modules/processHandling/processHandling.js
JavierJF: Because the command could fail in the first attempt, in the case of 'ClassicShell' for example we observed a delay of 6 seconds (depending on the system) while the daemon is not responsive to client petitions.

So a retry looked like the most logical thing to do in such cases when you can not expect responsiveness from external commands.

gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
```diff
@@ -260,6 +315,96 @@ gpii.windows.spi.populateResults = function (payload, isNewValue, isGetting, res
...
+
+    return s.split(" ").length;
+};
+
+/**
+ * @param {*} input String to be splitted into words.
+ */
+function splitInWords(input) {
+    var result = input.split(/[ ]+/);
+    if (result[(result.length - 1)] === "") {
```
stegru: `input.trim().split()`

gpii/node_modules/processHandling/processHandling.js
stegru: Yes. The purpose of closeApplications was to simulate a logout/in.

I'll be changing this stuff soon anyway, with the refactoring of GPII-2578.

gpii/node_modules/processHandling/processHandling.js
stegru: Yes, find a better name.

I'm not convinced this stuff belongs in a single function, but it works around the fact that the solutions registry doesn't "chain" multiple commands. Maybe the work @kaspermarkus is doing fixes this.

gpii/node_modules/processHandling/processHandling.js
stegru: Just 1 flag is needed:

```javascript
if (options.waitForClose) {
 // wait for close
} else {
 // wait for start
}
```
(or name it `waitForOpen` if the default of closing makes more sense)

gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
```diff
@@ -76,20 +79,31 @@ gpii.windows.spi.getCurrentSettings = function (payload) {
...
+    } else {
+        currentSettings.pvParam = pvParam.deref();
+    }
 
     return currentSettings;
 };
 
-gpii.windows.spi.systemParametersInfo = function (pvParamType, action, uiParam, pvParam) {
+gpii.windows.spi.systemParametersInfo = function (pvParamType, action, uiParam, pvParam, fWinIni) {
+    // Linter Workaround: Linter does not accept default parameters from ES6 spec.
```
stegru: So what?

gpii/node_modules/processHandling/processHandling.js
JavierJF: Yes yes, I knew, I wasn't just convinced that *the opposite* of waitForClose, being waitForOpen was natural, or evident. And that is why in first instance I chose strings. But it seems it looks evident to you, so I will go with the flag.

gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
stegru: `fWinIni = fWinIni || 0` - but I don't think it's needed, ffi will assume undefined is 0.

gpii/node_modules/processHandling/processHandling.js
JavierJF: I'm not also, this would sound more natural *as a composition of functions*, the problem as you pointed out, is that we need a entre point for the solution registry.  So I will try to find the *less ugly | more descriptive name* for this.

provisioning/data/scripts/TabletView.ahk
```diff
@@ -0,0 +1,23 @@
+SetBatchLines -1
+ListLines Off
+
+SELECTED_MODE = %1%
+TABLETMODESTATE_DESKTOPMODE := 0x0
+TABLETMODESTATE_TABLETMODE := 0x1
+
```
stegru: Interesting hack, but I think C# is more appropriate.

gpii/node_modules/processHandling/processHandling.js
JavierJF: I can extract in small function and make the string completion with parameters if needed.

gpii/node_modules/processHandling/processHandling.js
JavierJF: I like the idea, should be good to have a style-guide to decide this kind of situations for us, I use to trait function objects in a more classic way. I will try to minimize code with this kind of refactoring since now.

gpii/node_modules/processHandling/processHandling.js
JavierJF: I know, that was just because I though it was more clear to follow it in that way.

gpii/node_modules/processHandling/test/testProcessHandling.js
JavierJF: Okay

provisioning/data/scripts/ImmersiveShellRecover.ps1
```diff
@@ -0,0 +1,13 @@
+<#
```
stegru: Do these scripts only get ran in our vagrant box? Or on the client?

My understanding of the `provisioning` directory is it contains scripts to provision our VM. They don't get put into the installer. Why not have a top-level directory called `data`?

provisioning/data/scripts/StartMenuRecover.ps1
```diff
@@ -0,0 +1,49 @@
...
+  clear-host
+} else {
+  # We are not running "as Administrator" - so relaunch as administrator
+
+  # Create a new process object that starts PowerShell
+  $newProcess = new-object System.Diagnostics.ProcessStartInfo "PowerShell";
+  # Specify the current script path and name as a parameter
+  $newProcess.Arguments = $myInvocation.MyCommand.Definition;
+  # Indicate that the process should be elevated
+  $newProcess.Verb = "runas";
```
stegru: I know you know this: Either the UAC prompt will appear when keying in/out, or you'll have to disable the prompt. It's ok for a demo.

A better approach would be using this https://gist.github.com/stegru/7c8126259af943c4f8412710c2b53367 during install time (ran by the installer, as admin). This reduces the need to be administrator to change some registry values.

The best approach, however, is doing this kind of thing through a service (GPII-2338).

gpii/node_modules/processHandling/processHandling.js
stegru: Well, it's mainly for reducing duplicate code, rather than minimising it.

Another option is putting the return promise into a variable, and call .then on it once after.

gpii/node_modules/processHandling/processHandling.js
stegru: fine, I don't care.

gpii/node_modules/processHandling/test/testProcessHandling.js
JavierJF: okay, I see what you mean. Yes it could avoid the repetition, but it was for a test, so I didn't care. I will change it.

gpii/node_modules/WindowsUtilities/WindowsUtilities.js
JavierJF: Removing it, was making checks for the function not taking into account the use case. Bad habit, we have talked about it before.

gpii/node_modules/WindowsUtilities/WindowsUtilities.js
JavierJF: Yes, same as before.

gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
JavierJF: Well, having that type will make things easier for the future for sure, in case we need to reuse it. I will look into it and comment the results to you.

gpii/node_modules/spiSettingsHandler/src/SpiSettingsHandler.js
JavierJF: You are right, was added but never removed.

provisioning/data/scripts/TabletView.ahk
JavierJF: I also think it would be more appropriate, also, we have new tools that could give us much more better approaches than this, but the functionality was working and hasn't been touched since it was implemented. I will take a look to it and see how to create a better version for when this needs to reach master.

gpii/node_modules/processHandling/test/testProcessHandling.js
amb26: :+1: Test cases are code too :)

JavierJF: Haha sure! I'm working into all suggestions.

provisioning/data/scripts/StartMenuRecover.ps1
JavierJF: Yes, this need to change. As you have probably saw, the rest of the script just changes some registry keys and restart a process, so probably this all needs to be rewrited when [this pull](https://github.com/GPII/windows/pull/120) enter master.

provisioning/data/scripts/ImmersiveShellRecover.ps1
JavierJF: Yes, you are right, these scripts are copied during provisioning to the GPII data folder, which is the place we decided to keep for the scripts that will be later call in the system, if we don't have other working technique for doing the task.

gpii/node_modules/processHandling/processHandling.js
JavierJF: I have been seen not expected behavior when I have replaced this implementation with the suggested one using `closeProcessByName`, I think the reason is that `closeProcessByName` doesn't restart the process as the other function is supposed to do. I believe that the combination of registering the application for restart, and calling 'closeProcessByName' leads to process killing, and the restart never happens.

PR #158, opened 2017-12-08
GPII-2729: Onboard Windows 10 Internationalization settings

JavierJF:


kaspermarkus:
@JavierJF will notify @stegru when this is ready for review... He'll give it a first round of review, and pass to antranig when ready for merge/second round of review

gpii/node_modules/registrySettingsHandler/src/RegistrySettingsHandler.js
```diff
@@ -348,6 +348,32 @@ windows.enumRegistryValues = function (baseKey, path) {
     return togo;
 };
 
+windows.isSupportedMsg = function(message) {
```
stegru: I like the idea of being able to send a message. However, I think it should be placed into a function and called in a "start" or "stop" block, like `gpii.launch.exec` or `gpii.windows.closeProcessByName`.

That way, other settings handlers can make use of it.

gpii/node_modules/registrySettingsHandler/src/RegistrySettingsHandler.js
```diff
@@ -348,6 +348,32 @@ windows.enumRegistryValues = function (baseKey, path) {
...
+    var hwnd = windows.API_constants[message.hwnd];
```
stegru: Why bother? It's not like there can be other values for the hWnd.

I suggest either remove the ability to specify the hWnd, or perhaps accept a string, where it sends the message to windows with a matching caption (and/or classname?), if it's not "HWND_BROADCAST" (using a combination of `windows.enumerateWindows` and `GetWindowText`).

gpii/node_modules/registrySettingsHandler/src/RegistrySettingsHandler.js
```diff
@@ -348,6 +348,32 @@ windows.enumRegistryValues = function (baseKey, path) {
...
+    var msg  = windows.API_constants[message.msg];
+
+    return hwnd !== undefined && msg !== undefined;
+}
+
+windows.sendWindowMessage = function(message) {
+    try {
+        if (!windows.isSupportedMsg(message)) {
```
stegru: Why not just do `!hwnd || !msg` on 367

stegru:
@JavierJF, that's my first-pass review. I haven't ran it (it has the "Needs Work" label).

I agree with your approach. There is a [SetLocalInfo](https://msdn.microsoft.com/library/dd374049.aspx) API function, and I think that does exactly what you've done - I mean, this works, right? It might be worth creating a small .c program that just calls that function to see if it does anything else... but I doubt it does.

I also have an idea - perhaps there could also be a "culture" setting (en-GB, es-ES). It would make all of these individual settings whatever that culture default is for that setting, unless specifically set. [GetLocalInfoEx](https://msdn.microsoft.com/library/dd318103.aspx) gets this information.

gpii/node_modules/registrySettingsHandler/src/RegistrySettingsHandler.js
JavierJF: I like that thought, will be simpler, and useful from all the SettingsHandlers, sadly we have a imminent change for that block coming from [this pull](https://github.com/GPII/windows/pull/120). So probably I'm going to update the rest of issues of the pull and leave this one as a **TODO**, to be fixed when the new blocks could be used.

stegru:
ok to test

stegru:
What changed since I last looked?

My suggestions:
- The registry handler isn't the best place to send messages from. There's now a windowMessages module, https://github.com/gpii/windows/blob/master/gpii/node_modules/windowMessages/src/windowMessages.js (it currently just receives messages)
- Use the same mechanism as `gpii.launch.exec` or `gpii.windows.closeProcessByName`, so other solutions without registry settings can enjoy sending messages.
- No need to pass `HWND_BROADCAST` (just always use it), because there is no other _valid_ window handle constant that I'm aware of. Bonus points: allow passing the window class name.
- Maybe put the WM_xx constants in their own object.

stegru:
ok to test

stegru:
ok to test
```
    default: WinRM transport: negotiate
An error occurred executing a remote WinRM command.

Shell: Cmd
Command: hostname
Message: Unable to parse authorization header. Headers: {}
Body: SSH-2.0-OpenSSH_7.6
Protocol mismatch.
 ().
Build step 'Execute shell' marked build as failure
Finished: FAILURE
```

gpii/node_modules/windowMessages/src/windowMessages.js
```diff
@@ -238,3 +238,34 @@ gpii.windows.messages.destroyMessageWindow = function (that) {
 };
 
 gpii.windows.messages();
+
+/**
+ * Sends a message to every open window.
+ *
+ * @param message.msg {number} The message to be sent.
```
stegru: No doc for the actual `message` parameter.

In fact, why not unroll these into individual parameters, since they're all required?

gpii/node_modules/registrySettingsHandler/src/RegistrySettingsHandler.js
```diff
@@ -370,6 +370,7 @@ windows.registrySettingsHandler.setImpl = function (payload) {
             newValue: newValue.value
         };
     });
+
```
stegru: Stray newline

gpii/node_modules/windowMessages/src/windowMessages.js
```diff
@@ -238,3 +238,34 @@ gpii.windows.messages.destroyMessageWindow = function (that) {
...
+ * @param message.wparam {Object} The wparam to be passed to SendWindowNotifyMessageW function.
+ * @param message.lparam {Object} The lparam to be passed to SendWindowNotifyMessageW function.
```
stegru: `wparam` should be `{Number}`, and `lparam` a `{String}` (considering line 255)

PR #159, opened 2017-12-08
GPII-2751: Gibberish from the registry

stegru:
Added null characters to values read and written to the registry.

See [GPII-2751](https://issues.gpii.net/browse/GPII-2751?focusedCommentId=29806&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-29806).

### DIFF
--- a/gpii/node_modules/registrySettingsHandler/test/testRegistrySettingsHandler.js
+++ b/gpii/node_modules/registrySettingsHandler/test/testRegistrySettingsHandler.js
@@ -216,7 +216,7 @@ gpii.tests.windows.registrySettingsHandler.enumerationTest = {
         },
         "ZeroLength": {
             "data": "",
-            "bytes": 0,
+            "bytes": 2,
             "name": "ZeroLength",
             "type": "REG_SZ"
         }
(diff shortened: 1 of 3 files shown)


stegru:
oh @javihernandez, this fixes the "elf characters"

gpii/node_modules/registrySettingsHandler/test/testRegistrySettingsHandler.js
```diff
@@ -275,13 +283,26 @@ jqUnit.test("Reading and writing single registry keys", function () {
...
 
+    // Test zero-length values return an empty string (GPII-2751).
+    // The failure of this test also depends on the 2 bytes after the buffer being zero, so run it a few times to reduce
+    // the element of luck.
+    for (var retry = 0; retry < 50; retry++) {
+        var emptyString = "";
+        gpii.windows.writeRegistryKey("HKEY_CURRENT_USER", "Software\\GPIIMagnifier", "ZeroTest", emptyString, "REG_SZ");
+        var value5 = gpii.windows.readRegistryKey("HKEY_CURRENT_USER", "Software\\GPIIMagnifier", "ZeroTest", "REG_SZ");
+        jqUnit.assertEquals("Assert empty REG_SZ", 0, value5.value.length);
+        if (value5.value.length) {
```
kaspermarkus: Silly detail, but perhaps explicitly put `values5.value.length !== 0` - just to make it clear that we're interested in the fact that it's 0 or not

Also, could you rename value5 to something else, else idiots like me will start trying to figure out what the 5 refers... valueRead, registryValue, or the like would probably be more clear

stegru: So that test depends on some arbitrary byte (which we have no control over) not being `0`, and that's the best you can come up with? ;-)

kaspermarkus:
Talked with @javihernandez about the status of this. Since he was able to reproduce the original bug, he will test if this code fixes it, and merge it if that's the case

PR #161, opened 2018-02-09
GPII-2568: Updated references to universal with gpii-universal

javihernandez:
Works with https://github.com/GPII/universal/pull/548


javihernandez:
ok to test

PR #165, opened 2018-02-27
GPII-2332: Merge relevant stuff from hst-2017 into master

javihernandez:
I think it's a good time to completely get rid of this branch.
This pull request is not meant to be merged as it is, so I will create a separate one when we agreed on what we really want "in".

Actually, this branch includes:

1. Updated provisioning scripts. We want these as they are.
2. The VAIO proximity listener. We can skip this one since the merge of the in-process user listeners is going to happen right after the node upgrade.
3. [GPII-2200](https://issues.gpii.net/browse/GPII-2200), the debug logger. If memory serves me well, we decided that we wanted to get this merged into master, but I'd like a double check.

@amb26, @stegru does this sound good to you?


javihernandez:
This is the small pr for the provisioning stuff https://github.com/GPII/windows/pull/166

amb26:
Could you remind us here what else we require from this branch in addition to #166 ?

javihernandez:
> Could you remind us here what else we require from this branch in addition to #166 ?

The bits for GPII-2200, I can prepare another pull request to include them too.

stegru:
Javi, I'd feel better if I re-visited GPII-2200. It was only intended for HST, and the logging may be too invasive for general use. In fact, there hasn't been a time that information was required (it was there just in case).

Also, I don't think the code is actually called (it's just there), so it's not as though anyone would benefit from it at the moment.

amb26:
Looks dead

PR #166, opened 2018-03-01
GPII-2884: Moving provisioning scripts from hst-2017 into master

javihernandez:


amb26:
Hi there @javihernandez - could you merge this up with the updated master and verify that running "npm install" now builds the file gpii/node_modules/processHandling/test/test-window.exe - cheers!

stegru:
@javihernandez, test-window doesn't require anything more than what the native add-ons require. The hard-coded path is embarrassing, however I'm unaware of a better way to load "vsbuildtools.bat" (which fills the environment with paths to the toolchain and sdk). It does assume it's running on a 64-bit system.

Workarounds include:

- Creating a .msbuild file (and perhaps a .vcxproj), and running msbuild on it. This is what node-gyp does to build the native add-ons.
- Write it in C#, the path to the compiler is easier to find.
- Put the binary in git.
- Minimal solution: detect 32/64 bit and use the appropriate program files environment variable.
- Something else.

The only purpose of this file is to get some code running in a process that isn't named "node.exe", in order to test killing a process by name.

Also, you might want to see this if you haven't already: https://github.com/GPII/windows/pull/167/commits/c6115a7ef179a7a978beae306c2021c11350eabc#diff-8b49d7554433c0cb2f8d728d3135025dR19

javihernandez:
@stegru thanks for commenting.
Yes, in fact we don't need anything else, it's already there. The thing is that right now we depend on stuff from MS to perform the build, which is being installed on our vagrant boxes through Chocolatey, and I guess that we want to revisit [this](https://github.com/idi-ops/packer-windows/blob/master/scripts/chocolatey-packages.bat#L11-L14) and [this](https://github.com/GPII/windows/blob/master/provisioning/Chocolatey.ps1#L19-L22). As I commented earlier, I guess that we can get rid of some of this stuff and also, to rely on https://www.npmjs.com/package/windows-build-tools, which simplifies the things a bit since it sets up the system and looks like it's the most comfortable way for setting up your machine to build native add-ons (at least for someone who is not very familiar with setting up a windows machine to build stuff). Also, I think that our DevOps would give us a thumbs up since this approach can reduce the size of our windows virtual machines.

I didn't know that you were already moving this into an npm script on your pull request for GPII-2772. In fact, we're doing exactly the same. If you want, we can pair on this and come up with the best strategy.

stegru:
Well, the best strategy in my opinion is to change how test-window.exe is built - it's caused more torment integrating it into the build than it did producing it... and it's only ran in the unit tests (to its credit, it has recently revealed a bug). It will probably take less time re-writing it in C# than it would figuring out a gracious way of building the current one. That way it wouldn't require any special build libraries at all.

And as far as I know, the native node add-ons just need windows-build-tools (which includes its own python), and the C# stuff is built using .NET which is already on the machine. I don't know about the installer, however.

stegru:
True to my word; https://github.com/stegru/windows/commit/14ca491c0feecffd02817e590a2bb9dd624f453d

javihernandez:
> True to my word; stegru/windows@14ca491

Good stuff Mr Grudell!! @amb26 how do you want to proceed?

PR #169, opened 2018-03-20
GPII-2578: Make more use of the process reporter

stegru:
There's some overlap with processHandling and processReporter. This PR removes functions from processHandling that have been implemented in processReporter.

processHandling now (mostly) just contains process _actions_.


gpii/node_modules/processHandling/processHandling.js
```diff
@@ -24,22 +24,25 @@ var gpii = fluid.registerNamespace("gpii");
...
  * "firefox.exe".
  */
 gpii.windows.killProcessByName = function (filename) {
-    var pids = gpii.windows.findProcessByName(filename, true);
-    if (pids) {
-        for (var n = 0, len = pids.length; n < len; n++) {
-            var hProcess = windows.kernel32.OpenProcess(c.PROCESS_TERMINATE, 0, pids[n]);
+    var procs = processesBridge.findProcessesByCommand(filename);
+
+    if (procs) {
```
klown: `findProcessesByCommand()` always returns an array, even if it's empty.  No need to check if `procs` is non-null

gpii/node_modules/processReporter/dotNetProcesses.csx
```diff
@@ -51,7 +51,8 @@ public class Startup
                 }
             } else {
                 ManagementObject[] someProcesses = Array.FindAll(
-                    processes, p => p.GetPropertyValue("Name").ToString() == input
+                    processes,
+                    p => string.Equals(p.GetPropertyValue("Name").ToString(), input, StringComparison.OrdinalIgnoreCase)
```
klown: Thanks for catching this.  I had no idea that Windows processes were case insensitive.  Live and learn.

gpii/node_modules/processReporter/processesBridge.js
```diff
@@ -35,6 +35,11 @@ fluid.defaults("gpii.processes.windows", {
             funcName: "gpii.processes.windows.getProcessList",
             args: ["{arguments}.0"]
                    // process name or id, optional.
+        },
+        compareProcessNames: {
```
klown: I'm wondering if this belongs upstream in universal as part of the `"gpii.processes"` base grade.  I don't see anything Windows specific about it.  The implementation uses the standard JavaScript `String.toLower()` function.  Is there anything OS specific here?

gpii/node_modules/processReporter/test/processesBridge_tests.js
```diff
@@ -129,6 +129,25 @@ jqUnit.test(
...
+        var nodeProcInfos = processesBridge.findProcessesByCommand("NoDE.eXe");
+        jqUnit.assertNotEquals(
+            "Getting all 'NoDE.eXe' processes", 0, nodeProcInfos.length
+        );
+        nodeProcInfos.forEach(function (aProcInfo) {
+            jqUnit.assertEquals(
+                "Node commmand name", "node.exe", aProcInfo.command
+            );
+        });
+        var procInfo = processesBridge.findFirstProcessByCommand("NoDE.eXe");
```
klown: Love it!  Someone hiccoughed typing "nOde.ExE".  :-)

gpii/node_modules/windowsMetrics/test/WindowsMetricsTests.js
```diff
@@ -692,13 +692,17 @@ jqUnit.asyncTest("Testing application metrics", function () {
...
     // Pick three windows (owned by different processes) that already exist, and pretend they've became active by
     // mocking GetForegroundWindow.
     var windows = [];
     var exes = {};
     gpii.windows.enumerateWindows(function (hwnd) {
         if (windows.length < 3) {
-            var exe = gpii.windows.getProcessPath(gpii.windows.getWindowProcessId(hwnd));
+
+            var proc = processesBridge.findProcessByPid(gpii.windows.getWindowProcessId(hwnd));
+            var exe = proc && proc.fullPath;
```
klown: After seeing at least two occurrences of `proc && proc.fullPath`, my gut says to add a `getProcessPath()` invoker to the base grade in universal.

gpii/node_modules/processHandling/processHandling.js
stegru: as one would expect.

gpii/node_modules/processReporter/dotNetProcesses.csx
stegru: For some unknown reason, the `test-window.exe` process was being reported as `TEST-WINDOW.EXE`.

gpii/node_modules/processReporter/processesBridge.js
stegru: I was on the fence with this.

My reason was the code isn't windows-specific, but what it does is specific. But, it does seem to be a lot of "fluff" just for a `===` statement - especially considering the "real" comparison is in the .csx file.

I've moved the actual checking into universal, but left the decision of ignoring the case in windows.

gpii/node_modules/windowsMetrics/test/WindowsMetricsTests.js
stegru: My gut agreed.

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -24,22 +24,25 @@ var gpii = fluid.registerNamespace("gpii");
...
+        for (var n = 0, len = procs.length; n < len; n++) {
```
klown: I should have thought of this before, but ... using an iterator like [`fluid.each()`](https://docs.fluidproject.org/infusion/development/CoreAPI.html#fluideachsource-func) here would be more transparent.

gpii/node_modules/processReporter/processesBridge.js
klown: That makes sense.

The macOS file system can be either case insensitive or case sensitive depending on how it's configured (factory setting is insensitive).  I don't know about Android or iOS, but I suspect the former is sensitive and the latter is insensitive.  This is an invoker that's context dependent (infusion-speak).

gpii/node_modules/windowsMetrics/test/WindowsMetricsTests.js
klown: Cool, I'll go give that pull request a peek.

PR #173, opened 2018-05-03
GPII-1230: Including changes from Joseph

kaspermarkus:


amb26:
I have merged this to allow progress in other parts of the project, but there remains a bug https://issues.gpii.net/browse/GPII-2992 where the focus tracking rectangle remains in cyan. I assume this is something to do with the new enableRegisteredAT launch system.

PR #179, opened 2018-06-01
GPII-3092: Removed installer

stegru:
Removed the installer - only the installer from gpii-app is required.


gpii/node_modules/registryResolver/test/testRegistryResolver.js
```diff
@@ -31,8 +31,8 @@ jqUnit.test("Test Boolean Registry Lookups", function () {
     jqUnit.expect(2);
 
     jqUnit.assertTrue("Testing a registry key that always exists.",
-        gpii.deviceReporter.registryKeyExists("HKEY_CURRENT_USER",
-            "Software\\Microsoft\\Command Processor", "CompletionChar", "REG_DWORD"));
+        gpii.deviceReporter.registryKeyExists("HKEY_LOCAL_MACHINE",
```
amb26: Has this picked up something from some ancient/stray branch? The change looks reasonable, but why has it showed up in this diff?

stegru: The removed key was added by the Wix installer compiler, but now that wix has been removed, this key is no longer being created.

amb26: Why did the Wix installer compiler tinker with the command processor completion char?

stegru: oh I'm talking about the installer *for* WiX - the installation process of the Wix toolset.

PR #180, opened 2018-06-07
GPII-3099: Made GPII "DPI Aware"

stegru:
Calling [SetProcessDpiAwareness](https://msdn.microsoft.com/library/dn302122) so Windows doesn't lie about the non-client metrics.


amb26:
Astonishing and fantastic pull

PR #167, opened 2018-03-07
GPII-2772: Windows system settings handler

stegru:
This is the settings handler that lets you perform changes using the same back-end as the system settings applet (aka, immersive control panel).

The main piece of work is performed in an external helper C# program, because it has to be running as 64-bit. This is built during `npm install` (via [provisioning/NpmInstall.ps1](https://github.com/stegru/windows/blob/GPII-2772/provisioning/NpmInstall.ps1#L17)).

PR #183, opened 2018-07-10
GPII-3140: Made enableRegisteredAT.set return a promise

stegru:
Fixes intermittent failures in the acceptance tests.

Like https://github.com/GPII/windows/pull/167#issuecomment-400005673


stegru:
ok to test (for more confidence)

PR #185, opened 2018-07-19
GPII-3102: Metrics QSS support

stegru:
The windows portion of https://github.com/GPII/universal/pull/632.


stegru:
ok to test, installer is now gone.. https://github.com/GPII/windows/pull/179

PR #192, opened 2018-08-12
NOJIRA: Restarting explorer manually if RmRestart fails.

stegru:
Fixes `15:53:07.541:  FATAL ERROR: Uncaught exception: win32 error: RmRestart return:352 win32:1314`

### DIFF
--- a/gpii/node_modules/processHandling/processHandling.js
+++ b/gpii/node_modules/processHandling/processHandling.js
@@ -18,7 +18,8 @@
 "use strict";
 
 var fluid = require("gpii-universal");
-var ref = require("ref");
+var ref = require("ref"),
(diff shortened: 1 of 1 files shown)


stegru:
ok to test

amb26:
Is this a fix for the issue Gregg reported at https://lists.gpii.net/pipermail/architecture/2018-August/004878.html or did this already get into feds_audit and he has a further cause of the issue?

stegru:
This fix, and https://github.com/GPII/windows/pull/191 should have worked. They where in the build that Gregg was using (right, @javihernandez?)

Either they don't work, or there's another cause.

Let's ditch this one, and I'll combine this into an improved https://github.com/GPII/windows/pull/191

PR #194, opened 2018-08-27
GPII-3305: Improved explorer restart.

stegru:
Now starts and stops explorer without using the [Restart Manager API](https://docs.microsoft.com/windows/desktop/RstMgr/restart-manager-portal).

The api was originally used because, at first, other applications where being restarted. However, this has been proven to be a bad idea. (While Kavya was demonstrating the language feature during Toronto 2018 F2F, it killed her Vidyo process).

So, it's not only over-kill for a single process, but also a waste of time because due to its unreliability, some "manual" restarting code also had to be implemented on top of it.

Explorer is now gracefully closed, and restarted manually. Explorer still gets in a state where it doesn't completely shutdown - in that instance, it's killed when the shutdown times out.


gpii/node_modules/processHandling/processHandling.js
```diff
@@ -288,112 +289,61 @@ gpii.windows.getServiceState = function (serviceName) {
...
-gpii.windows.currentRestartSession = null;
-
 /**
- * Stops the Windows Explorer shell process, in a way that it can be restarted again so the folder windows are restored.
+ * Stops the Windows Explorer shell process.
  *
  * Use restartExplorer() to restart it.
  *
+ * @param {Object} options [optional] Options object.
+ * @param {Number} options.trayWindow Override the handle of the task tray window.
```
amb26: Say what kind of handle it is - process handle or window handle?

gpii/node_modules/processHandling/processHandling.js
```diff
@@ -288,112 +289,61 @@ gpii.windows.getServiceState = function (serviceName) {
...
+ * @param {Number} options.timeout Timeout, in ms, to wait before forcefully terminating the process.
```
amb26: Say the default

gpii/node_modules/processHandling/processHandling.js
stegru: done

gpii/node_modules/processHandling/processHandling.js
stegru: "the default"

PR #200, opened 2018-09-18
GPII-3371: Restoring cursor position back to the physical position before resolution/DPI change.

stegru:
When changing the DPI scale or screen resolution, the cursor is moved to the centre of the screen (this does not happen in a virtual machine, because the host owns the pointer).

(M3: https://docs.google.com/document/d/1NODvtQA6YKRdIsIDSLP-gqcnN6nTkxg8cqlPoNCGTDQ/edit)

This fix moves the mouse cursor back to the physical position before the screen change. That is, if you put your finger on the mouse pointer before changing the DPI/resolution, the pointer should remain under your finger (if your finger moves, there may be medication to prevent this).

**How to test**

* If running in a VM, uncheck *Mouse integration* from the *Input* menu (after making sure you know the key used to release the mouse - usually right ctrl). This is to allow the guest to move the mouse pointer.
* For testing the resolution, enable *Scaled mode* from the *View* menu (`Host`+`C`) in order to simulate a physical screen that doesn't change its size.

`elmer` changes the resolution, `salem` does the scale (currently broken: https://github.com/GPII/universal/pull/643).

### DIFF
--- a/gpii/node_modules/WindowsUtilities/WindowsUtilities.js
+++ b/gpii/node_modules/WindowsUtilities/WindowsUtilities.js
@@ -267,6 +267,10 @@ windows.user32 = ffi.Library("user32", {
     // https://msdn.microsoft.com/library/ms633499
     "FindWindowW": [
         gpii.windows.types.HANDLE, ["char*", "char*"]
+    ],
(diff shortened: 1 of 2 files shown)


amb26:
Good job!

PR #201, opened 2018-09-19
GPII-3349: Metrics improvements

stegru:
This adds the following new metrics:
* System info (CPU, Memory, display, windows version)
* GPII versions
* Mouse wheel
* A README file
* The ability to disable mouse/key metrics (eg, if anti-virus complains, or it's too much data).

Updates the following:
* Window activation:
  * logged when it happens, removed duration active (which can be implied).
  * Added window identifier.
  * Using WM_SHELLHOOK notification, rather than polling.
* Modifier key state during key/mouse input.

(no changes to universal)


stegru:
ok to test

stegru:
ok to test

PR #208, opened 2018-10-12
GPII-3438: Update to Node 10 +ffi replacement with ffi-napi

javihernandez:


gpii/node_modules/windowMessages/src/windowMessages.js
```diff
@@ -200,15 +200,16 @@ gpii.windows.messages.createMessageWindow = function (that) {
         cls.lpszClassName = className;
 
         // Create a pointer to the window procedure function.
-        cls.lpfnWndProc = ffi.Callback(gpii.windows.types.HANDLE,
+        var callback = ffi.Callback(gpii.windows.types.HANDLE,
```
amb26: Any reason for this diff?

javihernandez: I think it's just for readability reasons, but I'm not sure about the reason for this: https://github.com/GPII/windows/pull/208/files#diff-6dcc803c806d1e5abf989cedcb2b735cR208
@stegru, is that because cls.lpfnWndProc needs to hold a reference to the callback after all?

javihernandez:
> CI job failed: https://ci.gpii.net/job/windows-tests/686/

Here's the problem:
```
..\src\binding.cc(5): fatal error C1083: Cannot open include file: 'napi.h': No such file or directory [C:\vagrant\node_modules\ref-napi\build\binding.vcxproj]
VERBOSE: 'npm' exited with 1
```
It happens that if I run **npm install** from *v:\ * (of victory) this problem is not present. :man_shrugging:

gpii/node_modules/windowMessages/src/windowMessages.js
stegru: yes @javihernandez - the callback gets called throughout the lifetime of the process. This is to stop it being garbage collected when the function exits.

amb26: Surely it shouldn't make a difference to its GC status whether the global windowProcPointer references it as a member of cls or a top-level closure variable

amb26: Or is it that we're happy to have cls GCed but not the callback?

stegru: Right. cls is not required - nor is there any harm in making it global.

Before today, it was `cls` that was being kept warm (wasn't my intention, but did no harm other than wasting a handful of bytes).

stegru: There are two things that need to be referenced, the return of `ffi.Callback` (the function pointer), and the `cif` thing that's internal to ffi.

amb26: OK, could please add a comment explaining this subtlety where you write the callback, and also a link to the JIRA

amb26:
This now also needs a update with master

amb26:
Really, really great work guys! :tada:

PR #220, opened 2018-11-21
GPII-3544: Get installed languages.

stegru:
The detection of languages installed on the OS.

PR #221, opened 2018-11-30
GPII-3544 on 0.3.0-dev.20181120T145240Z.c46b4be: OS Languages

stegru:
So the languages can be implemented before upgrading electron/node.

https://github.com/GPII/windows/pull/220 is the real PR


stegru:
no longer required

PR #223, opened 2018-12-02
GPII-3544: Get installed languages (fix).

stegru:
This is a fix to https://github.com/GPII/windows/pull/220 (merged), which gets the list of languages from a better location.

It was returning just the language where the language is only used in a single country (or something similar to that effect). For example, `bg` was being returned, instead of `bg-BG`.

While `bg` is still a valid language code, it is not enough when setting the display language.


gpii/node_modules/gpii-localisation/src/language.js
```diff
@@ -112,103 +101,137 @@ gpii.windows.winlangdb = ffi.Library("winlangdb", {
...
-    var langCodes = gpii.windows.readRegistryKey(
-        "HKEY_CURRENT_USER", "Control Panel\\International\\User Profile", "Languages", "REG_MULTI_SZ").value;
-
-    var languages = {};
-    var count = 0;
+    var langCodes = gpii.windows.enumRegistryKeys(
+        "HKEY_LOCAL_MACHINE", "SYSTEM\\CurrentControlSet\\Control\\MUI\\UILanguages");
+
+    var knownLanguages = Object.keys(that.model.installedLanguages || {});
+    var changed = langCodes.length !== knownLanguages.length;
```
danailbd: Doesn't this check only whether there's a newly installed language?
Would the list of `langCodes` change in case the current language changes?

gpii/node_modules/gpii-localisation/src/language.js
```diff
@@ -112,103 +101,137 @@ gpii.windows.winlangdb = ffi.Library("winlangdb", {
...
-        languages[code] = that.getLanguageNames(code);
-        languages[code].code = code;
-        languages[code].current = code === current;
-        languages[code].order = count++;
-    });
-
-    that.applier.change("installedLanguages", languages);
-    that.applier.change("configuredLanguage", gpii.windows.language.getDisplayLanguage());
+            that.applier.change("installedLanguages", languages);
+            that.applier.change("configuredLanguage", current);
```
danailbd: Currently the `configuredLanguage` doesn't get changed so there might be a problem here. :/
Isn't there a case when the `configuredLanguage` changes without the `installedLangauges` getting changed?

gpii/node_modules/gpii-localisation/src/language.js
```diff
@@ -112,103 +77,140 @@ gpii.windows.winlangdb = ffi.Library("winlangdb", {
...
+        promise = fluid.promise().resolve();
+    }
 
-    fluid.each(langCodes, function (code) {
-        languages[code] = that.getLanguageNames(code);
-        languages[code].code = code;
-        languages[code].current = code === current;
-        languages[code].order = count++;
+    promise.then(function () {
+        that.applier.change("configuredLanguage", current);
```
danailbd: The language gets updated properly but the `installedLanguages`' `local` values don't :/
Probably whenever the language changes, the `installedLanguages` should get updated as well?
But there's also the case when we might change from `en-US` to `en-GB` where the `local` names of the language might not change...

gpii/node_modules/gpii-localisation/src/language.js
```diff
@@ -281,3 +283,5 @@ fluid.defaults("gpii.windows.updateLanguage", {
         currentLanguage: 0
     }
 });
+
+gpii.windows.language();
```
amb26: I don't think this file should automatically create a "floating component" immediately on creation. It's the integrator's responsibility to decide whether and where to create the component.

gpii/node_modules/gpii-localisation/src/language.js
amb26: :+1: - just apply the new list on top of the old one with the changeApplier. It will take care of diffing the old values with the new and deciding whether to issue an updated.

gpii/node_modules/gpii-localisation/src/languageNames.js
```diff
@@ -0,0 +1,139 @@
+/*
+ * Child process to get localised language names.
```
amb26: Refer to a JIRA or reproduce an explanation here about why this needs to be in a child process

gpii/node_modules/gpii-localisation/src/language.js
stegru: yes, I left that in by accident.

gpii/node_modules/gpii-localisation/src/language.js
stegru: The checks are just to avoid starting the child process for the translations.

New comment:
```
    // Because this function gets called in the off-chance that the language has changed (maybe multiple times during
    // key-in), and getting the translated languages requires a new process, perform some checks upfront.
```

gpii/node_modules/gpii-localisation/src/language.js
stegru: Fixed.

gpii/node_modules/gpii-localisation/src/language.js
stegru: Fixed

gpii/node_modules/gpii-localisation/src/languageNames.js
stegru: yes

stegru:
ok to test

amb26:
Got some peculiar test output from this in the VM - it took a huge amount of time (about 30 seconds) and then produced:
```
Bcp47GetXXXName failed (english) -2147024809
Bcp47GetXXXName failed (local) -2147024809
Bcp47GetXXXName failed (native) -2147024809
Bcp47GetXXXName failed (english) -2147024809
Bcp47GetXXXName failed (local) -2147024809
Bcp47GetXXXName failed (native) -2147024809
Bcp47GetXXXName failed (english) -2147024809
Bcp47GetXXXName failed (local) -2147024809
Bcp47GetXXXName failed (native) -2147024809
Bcp47GetXXXName failed (english) -2147024809
Bcp47GetXXXName failed (local) -2147024809
Bcp47GetXXXName failed (native) -2147024809
11:05:18.339:  jq: Test concluded - Module "gpii.tests
```

stegru:
Those errors are ok - they're from the "passing junk data" tests - however, 30 seconds isn't good. I can speed them up.

stegru:
On CI it was 12 seconds, now 2

ok to test

PR #228, opened 2018-12-13
GPII-3554: Add tests to wmiSettingsHandler to increase coverage

byoung2:


javihernandez:
ok to test

amb26:
ok to test

gpii/node_modules/wmiSettingsHandler/test/testWmiSettingsHandler.js
```diff
@@ -271,3 +271,53 @@ jqUnit.test("Testing wmiSettingsHandler set handles unsupported feature.", funct
...
+    jqUnit.expectFrameworkDiagnostic("Should fail due to bad getQuery", function() {
+        windows.wmi.getQuery (query);
+    }, 'Query failed with WMI error msg');
+});
+
+jqUnit.test("Testing wmiSettingsHandler updates successfully.", function () {
+    var brightnessVal = 30;
+    var payload = ["root\\WMI" ,"WmiMonitorBrightnessMethods", "WmiSetBrightness", [0xFFFFFFFF, brightnessVal], ["uint", 0]];
+
+    jqUnit.assert("Setting is supported in the system, so test passes with no validity.");
```
amb26: This test doesn't actually exercise the target code at all?

byoung2: There were 3 or 4 places that lacked test coverage. There is an added test to address each one.

amb26:
Cheers @byoung2 - looking great so far, good improvement in coverage. Looks like you didn't quite get the big function body at line 227 yet - see attached screenshot

gpii/node_modules/wmiSettingsHandler/test/testWmiSettingsHandler.js
```diff
@@ -272,6 +272,54 @@ jqUnit.test("Testing wmiSettingsHandler set handles unsupported feature.", funct
...
+                    "returnVal": [
+                        "uint",
+                        0
+                    ]
+                }
+            }
+        }
+    };
+    var mockWindows = windows;
+    mockWindows.wmi.getImpl = function (payload) {
```
amb26: The "mockWindows" variable doesn't add much since it just holds the same value as "windows". Instead you probably want to write something like 
````
var oldGetImpl = windows.wmi.getImpl;
try {
     windows.wmi.getImpl = .....
````
etc.

File: gpii/node_modules/wmiSettingsHandler/test/testWmiSettingsHandler.js [hunk 3]
```diff
@@ -272,6 +272,54 @@ jqUnit.test("Testing wmiSettingsHandler set handles unsupported feature.", funct
...
+            fluid.set(results, settingId + ".value", 20);
+        });
+
+        return results;
+    };
+
+    var result = mockWindows.wmi.setImpl(payload);
+    var oldValue = result.Brightness.oldValue === 20;
+    var correct = oldValue;
+    jqUnit.assertTrue("The handler should attempt to set value when setting is supported.", correct);
```
amb26: more idiomatically - 
    jqUnit.assertEquals("The handler .....", 20, result.Brightness.oldValue);

gpii/node_modules/wmiSettingsHandler/test/testWmiSettingsHandler.js
byoung2: I was just worried that if these tests ever were to run concurrently, my mocked test would affect other tests. But if that is not an issue your way is cleaner

byoung2: What is the purpose of the try block?

byoung2: I was thinking that one test could fire off an asynchronous operation, then the next test starts, then the first test calls back. I'll add the try catch finally block oer your suggestion.

amb26: Asynchronous tests always cause the underlying test engine (QUnit) to block so that their execution is serialised

byoung2: I added the try/catch/finally block. Makes more sense now. Thanks for the suggestion!

amb26:
Thanks for your great work on this, @byoung2 , and congrats on your first merge into the codebase!
:tada:

PR #230, opened 2018-12-28
GPII-3558: Closing explorer folder windows when restarting explorer.

stegru:
This should fix the issue where *explorer* is showing a mixture of languages. It ensures that all folder windows are also closed (they should also re-appear).

### DIFF
--- a/gpii/node_modules/windowMessages/src/windowMessages.js
+++ b/gpii/node_modules/windowMessages/src/windowMessages.js
@@ -255,4 +255,47 @@ gpii.windows.messages.destroyMessageWindow = function (that) {
     return success;
 };
 
+/**
+ * Send a message to a window, returning a promise which resolve when the message has been processed.
+ *
+ * @param {Number} hwnd The window handle.
+ * @param {Number} msg The message, or the name of a registered message (will be registered if not already).
+ * @param {Number} wParam Message parameter.
+ * @param {Number} lParam Message parameter.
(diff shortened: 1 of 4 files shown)


gpii/node_modules/windowMessages/src/windowMessages.js
```diff
@@ -255,4 +255,47 @@ gpii.windows.messages.destroyMessageWindow = function (that) {
...
+            if (timer) {
+                clearTimeout(timer);
+            }
+        });
+
+    gpii.windows.user32.SendMessageCallbackW(hwnd, msg, wParam, lParam, callback, 0);
+
+    // Keep the callback alive for as long as the promise.
+    promise.then(function () {
+        return callback && null;
```
amb26: Seems odd - given we have closed over it and we don't modify it, the callback could never be nulled out?

stegru: My intention isn't clear - that line is a no-op, that keeps the `callback` in scope.

Without it, `SendMessageCallbackW` tries to use it, sometimes after it's been GC'd.

amb26: It was clear in the end.... as I was noodling to myself about the code later on that day it suddenly struck me what it was for : P But I'm sure there's some other reader to whom it wouldn't be immediately obvious too - thanks for the comment!

PR #231, opened 2019-01-04
GPII-3130: Added the rest of settings required for Narrator to SR

JavierJF:


JavierJF:
Ok to test

settingsHelper/SettingsHelper/SettingHandler.cs
```diff
@@ -63,54 +63,103 @@ public static Result Apply(Payload payload)
...
+                    {
+                        // Only use those with the "Exposed" attribute.
+                        throw new SettingFailedException("Not an exposed method " + payload.Method);
+                    }
 
-                try
-                {
-                    result.ReturnValue = method.Invoke(settingItem, payload.Parameters);
-                    if (!(payload.Async ?? true))
+                    try
```
stegru: It's not immediately obvious what this block does, how about extracting it into its own method, with a nice comment?

settingsHelper/SettingsHelper/SettingHandler.cs
```diff
@@ -63,54 +63,103 @@ public static Result Apply(Payload payload)
...
-                        settingItem.WaitForCompletion(5);
+                        object expectedValue = null;
+                        if (payload.Parameters.Length == 1 && payload.Method == "SetValue")
+                        {
+                            Type paramType = paramTypes[0];
+                            expectedValue = Convert.ChangeType(payload.Parameters[0], paramType);
+                        }
+                        else
+                        {
+                            if (payload.ExpectedValue != null) {
```
stegru: This block will be called if `Method` isn't `Setvalue` - I don't think we'll need to get the result if it's another method.

settingsHelper/SettingsHelper/SettingHandler.cs
```diff
@@ -63,54 +63,103 @@ public static Result Apply(Payload payload)
 
                 payload.SettingItem = settingItem;
 
-                // Get the parameter types to get the right overload.
-                Type[] paramTypes = Type.EmptyTypes;
-                if (payload.Parameters != null)
-                {
-                    paramTypes = payload.Parameters.Select(p => p.GetType()).ToArray();
+                BindingFlags bindingFlags = BindingFlags.Instance | BindingFlags.Public | BindingFlags.IgnoreCase;
+                MethodInfo getMethod = settingItem.GetType().GetMethod("GetValue", bindingFlags, null, Type.EmptyTypes, null);
```
stegru: No need for reflection, `GetValue` can be called directly with `settingItem.GetValue`

settingsHelper/SettingsHelper/SettingHandler.cs
```diff
@@ -63,54 +63,103 @@ public static Result Apply(Payload payload)
...
+                            {
+                                throw new SettingFailedException("Unable to verify setting due to missing 'ExpectedValue' field in payload");
+                            }
+                        }
+
+                        if (!expectedValue.Equals(oldValue))
+                        {
+                            result.ReturnValue = method.Invoke(settingItem, payload.Parameters);
+
+                            if (!(payload.Async ?? true))
```
stegru: Here's an "is this setting async?" flag!

settingsHelper/SettingsHelper/SettingHandler.cs
stegru: Also, I think it can just have a `payload.CheckResult` boolean, instead of `expectedValue`. The expected value can always be taken from `payload.Parameters.Last()`

PR #232, opened 2019-01-07
GPII-3605: Correcting language code casing

stegru:
Corrects the casing of language codes, so `en-gb` becomes `en-GB`.


javihernandez:
LGTM 👍

PR #236, opened 2019-01-15
GPII-3596 Increase test coverage for appZoom

byoung2:


javihernandez:
ok to test

gpii/node_modules/gpii-app-zoom/test/testAppZoom.js
```diff
@@ -202,3 +202,216 @@ jqUnit.asyncTest("Send zoom", function () {
...
+        hwnd: appZoom.getMessageWindow(),
+        config: {
+            childWindow: "SimulatedChildWindow"
+        }
+    };
+    var oldGetClassNameW = gpii.windows.user32.GetClassNameW;
+    var oldStringFromWideChar = gpii.windows.stringFromWideChar;
+    var oldEnumerateWindows = gpii.windows.enumerateWindows;
+    try {
+        gpii.windows.user32.GetClassNameW = function () {
```
amb26: OK, I think we've got enough of these now that it is time to DRY up this pattern into a utility. Could you refactor these, as well as the existing tests you wrote for WMI, into a function that is something like this:
gpii.windows.test.testWithMocks({
    "gpii.windows.user32.GetClassNameW": true,
    "gpii.windows.stringFromWideChar": "SimulatedChildWindow",
    "gpii.windows.enumerateWindows": function (whnd, cb) {
            return cb("SimulatedChildWindow");
        }
    }, function () {
        appZoom.sendZoom("decrease");
        jqUnit.assertEquals("SimulatedChildWindow found", "SimulatedChildWindow", "SimulatedChildWindow");
    }, "Error zooming childWindow"
    };

gpii/node_modules/gpii-app-zoom/test/testAppZoom.js
```diff
@@ -202,3 +202,216 @@ jqUnit.asyncTest("Send zoom", function () {
...
+    appZoom.currentWindow = {
+        hwnd: appZoom.getMessageWindow(),
+        config: {
+            childWindow: "InvalidChildWindow"
+        }
+    };
+    var oldFluidLog = fluid.log;
+    var logMessage;
+    try {
+        fluid.log = function (msg) {
```
amb26: I think testing this facility via its side-effects on fluid.log is a bit unsatisfactory and we should instead improve the implementation so that it communicates this condition via a return value.

byoung2: Perhaps moving the check for childwindow into its own method with a return value of hwnd would be the best way. Let me do that

gpii/node_modules/gpii-app-zoom/test/testAppZoom.js
byoung2: Where is a good place for that utility to reside? The top level tests folder?

amb26: :+1:

gpii/node_modules/gpii-app-zoom/test/testAppZoom.js
byoung2: I pushed that change, now findChildWindow is a separate method with a hwnd return (the fluid.log is preserved)

amb26:
Some good improvements, but the majority of the fairly bulky function window.appZoom.findUncoveredPoint is uncovered (ha ha!). Can we break out the the interior of the function after if (rect) into a pure utility function and write some unit tests for that?
Also, window.appZoom.getConfig only documents the _window_ argument as being a WindowInfo - can we simply get rid of the untested "number" branch at line 396?

amb26:
Cheers Bryan, a good increment - the logic in the 2nd half of the function is still uncovered. It would be great if we could get this file, by hook or by crook, up somewhere near 80% branch coverage. Throwing some different varieties of rectangles, e.g. into getOverlappingPoints would also help to cover some of the untested "else" branches.
uncoveredpoint2

PR #243, opened 2019-02-01
 GPII-3644: Updating to latest universal release

amb26:


JavierJF:
Hi there @amb26 I think issue is that this branch is not updated against master, my current branch is, and that's why tests related to the "systemSettingsHandler" fails to set their preferences. Just updating this branch with master should make the tests pass. Thanks you!

PR #248, opened 2019-03-13
GPII-3739: Open USB folder.

stegru:
Back-end for the open usb button, which performs the following:

- If a single USB Drive is found, then that gets opened.
- If there's more than 1, then all drives that don't have a gpii token file are opened. (if all drives have a token then all are opened)

A "USB Drive" is a fixed or removal volume on the USB bus (which isn't a CD-ROM or Ramdisk).

Used by https://github.com/GPII/gpii-app/pull/92

### DIFF
--- a/gpii/node_modules/WindowsUtilities/WindowsUtilities.js
+++ b/gpii/node_modules/WindowsUtilities/WindowsUtilities.js
@@ -32,7 +32,9 @@ ffiBindings.Callback = function (cif, size, argc, errorReportCallback, fn) {
 
 var ffi = require("ffi-napi"),
     child_process = require("child_process"),
-    fluid = require("gpii-universal");
+    fluid = require("gpii-universal"),
+    fs = require("fs"),
+    path = require("path");
 
 var gpii = fluid.registerNamespace("gpii");
 var windows = fluid.registerNamespace("gpii.windows");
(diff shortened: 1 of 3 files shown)


stegru:
oops, it looks like I've included a commit from something else - unrelated to this, so I've undone it.

gpii/node_modules/WindowsUtilities/WindowsUtilities.js
```diff
@@ -1739,4 +1768,141 @@ gpii.windows.nativeExec = function (command, options, callback) {
...
+/**
+ * Used by the "Open USB" button. Gets the user's USB drives. That is, those without a token file.
+ * If all drives have a token file, then all are returned.
+ *
+ * @param {Array<String>} usbDrives [optional] The root paths to the usb drives to check, omit to check all USB drives
+ * on the system.
+ * @return {Array<String>} Array of paths to each drive.
+ */
+gpii.windows.getUserUsbDrives = function (usbDrives) {
+    usbDrives = usbDrives || gpii.windows.getUsbDrives();
```
amb26: Although there are test cases, coverage report shows the entirety of these two functions as uncovered..... odd?

amb26: My testing shows that the code is indeed hit, @the-t-in-rtf , can you comment?

amb26: Ah sorry, no mystery - WindowsUtilitiesTests.js are not included in the unit tests!

amb26:
Got a test failure in the vagrant box relating to order of the drives:
````
16:23:50.645:  jq: FAIL: Module "gpii.tests.windows.windowsUtilities" Test name "Testing getUserUsbDrives" - Message: getUserUsbDrives should resolve with the expected value - three drives, no token
16:23:50.645:  jq: Expected: [
    "A",
    "B",
    "C"
]
16:23:50.645:  jq: Actual: [
    "B",
    "C",
    "A"
]
````

gpii/node_modules/WindowsUtilities/WindowsUtilities.js
the-t-in-rtf: In addition, in anything that uses "node_modules" as a place to house code, there is always extra work to be done to keep nyc from ignoring it:

https://github.com/GPII/universal/blob/b4ef6b25c1586f7e79540e28ec04052d230bfb5a/.nycrc#L118

PR #255, opened 2019-03-25
NO-JIRA: Not listing USB removable drives (like SD cards).

stegru:
Fixes the problem where "phantom usb drives" where being shown with the open USB button.

These drives are probably empty SD cards.


stegru:
Closing: this make things worse.

PR #261, opened 2019-04-08
GPII-3844: Prevent system wallpaper corruption when setting high-contrast mode

JavierJF:


package.json
```diff
@@ -17,7 +17,7 @@
     "dependencies": {
         "edge-js": "10.3.1",
         "ffi-napi": "2.4.3",
-        "gpii-universal": "0.3.0-dev.20190329T151131Z.02584b73",
+        "gpii-universal": "JavierJF/universal#GPII-3844",
```
amb26: You will need to add rimraf as a direct dependency otherwise it will simply resolve "by accident" in npm's usual way

JavierJF: Right, I found that it was used in the "WindowsUtilitiesTests.js", so I followed touching the less I could. But, it's because it was in "devDependencies", so I will add it to "dependencies".

javihernandez:
this needs to be updated with current master

javihernandez:
@JavierJF can you update with current master?

JavierJF:
@javihernandez updated!

gpii/node_modules/WindowsUtilities/WindowsUtilities.js
```diff
@@ -1913,4 +1914,63 @@ gpii.windows.getUserUsbDrives = function (usbDrives) {
...
+                result.resolve();
+            } catch (err) {
+                result.reject({
+                    code: err.code,
+                    message: "Removing dir failed with error '" + err + "'"
+                });
+            }
+        } else {
+            try {
+                fs.unlinkSync(path);
```
javihernandez: We should resolve the promise after this is completed, right?

gpii/node_modules/WindowsUtilities/WindowsUtilities.js
```diff
@@ -1913,4 +1914,63 @@ gpii.windows.getUserUsbDrives = function (usbDrives) {
...
+        recursive: false,
+        silent: false
+    };
+    options = fluid.extend(defaultOptions, options);
+    var result = fluid.promise();
+
+    if (fs.existsSync(path)) {
+        if (options.recursive) {
+            try {
+                rimraf.sync(path);
```
javihernandez: I'd say that we should use the async form of rimraf here since this code can block the main thread for some time when processing a large directory.

stegru: I agree, however this is a solutions registry action which do not yet support async functions.

Perhaps make it async if it's ok for the files to still exist after the function returns (in other words, would things break if it was called at the end of the `configure` block).

javihernandez: Well spotted @stegru. I guess that we should live with the sync form for now. :/
@JavierJF can you add a TODO item here that indicates that we should move this to be async by when the support for async lifecycle actions is implemented? I found this relevant JIRA https://issues.gpii.net/browse/GPII-434

gpii/node_modules/WindowsUtilities/WindowsUtilities.js
JavierJF: Nice catch! Have been only tested for directories yet I think.

gpii/node_modules/WindowsUtilities/WindowsUtilities.js
JavierJF: @javihernandez sure, I will add the comment with that. Also, the operation needs to be sequential with the rest of the operations done int he configure block, so, in order to use the async version of this, we should support async lifecycle actions and being able to impose sequentiality between them. This last requirement is not described in the ticket, but I will update with a comment.

JavierJF:
Hi there @amb26 , could you give a quick look to this pull to see if it's ready? Thanks!

PR #262, opened 2019-04-12
GPII-3852: Stop showing languages that are removed.

stegru:
This fix stops the QSS from showing languages after they have been removed.

To test:

1. Install a language, start morphic: the language should be on the pop-up.
2. Remove the language, restart morphic: the language should no longer be seen.


javihernandez:
ok to test

PR #265, opened 2019-04-15
GPII-3866: Explorer wont restart in certain timing conditions

JavierJF:
This patch was suggested by @stegru to improve the order of promise resolution that guarantees a proper explorer shutdown.


stegru:
The returned promise was being resolved before explorer was fully killed, meaning explorer was getting started again too soon, causing it to get killed.

Now, the promise will only resolve after it's killed.

PR #266, opened 2019-05-21
GPII-3930: Create a "brightness" generic term

JavierJF:


JavierJF:
Closing this pull which only purpose was to make tests pass.

PR #269, opened 2019-06-21
GPII-3987: Expose current system brightness in isolated function

JavierJF:


gpii/node_modules/wmiSettingsHandler/src/WmiSettingsHandler.js
```diff
@@ -415,3 +415,33 @@ gpii.deviceReporter.wmiSettingSupported = function (namespace, query) {
 
     return result;
 };
+
+/**
+ * Gets the current system brightness value making the proper
```
amb26: Please add a comment referencing GPII-3987 and also that this function will be removed once GPII-3958 is implemented

gpii/node_modules/wmiSettingsHandler/src/WmiSettingsHandler.js
```diff
@@ -415,3 +415,33 @@ gpii.deviceReporter.wmiSettingSupported = function (namespace, query) {
...
+ * query to the wmiSettingsHandler.
+ *
+ * @return {Object} The current system brightness, if the
```
amb26: Is it an Object? It looks like it is a Number.... please also report what the scale of the value is

gpii/node_modules/wmiSettingsHandler/src/WmiSettingsHandler.js
JavierJF: Done!

gpii/node_modules/wmiSettingsHandler/src/WmiSettingsHandler.js
JavierJF: Yes, both observation were right, already done!

amb26:
Sorry to leave the review comments - I understood from our updates on Monday that this work would now no longer be required, since the next built of Morphic will not include the Brightness setting, and by the time we make the next build, @cindyli 's work on https://issues.gpii.net/browse/GPII-3958 will be complete.
Let me know if this memory/understanding of our status on Monday agrees with yours - if so, we can close this pull and the associated JIRA

JavierJF:
@amb26 yes, that was my understanding too, I just completed the feedback, just in case this works became relevant again before we have the other solution in place. It was just a 'anticipatory movement'. Now that it's addressed we can close everything since as you said, it looks like it wont be needed!

PR #271, opened 2019-07-05
GPII-4011: Showing the volume change on the screen

stegru:
This sets the volume to almost the required value, then uses the shell to perform the final increment causing the volume to show on screen.

The diff shown for [VolumeControl.cpp](https://github.com/GPII/windows/compare/master...stegru:GPII-4011?expand=1#diff-4b8f65b352dd4b891144769e3d5d6b84) is exaggerated, so here's a picture of the real changes:
VolumeControl.cpp


amb26:
This pull needs some dedicated unit tests for the newly added nativeSettingsHandler code

javihernandez:
@stegru, I see that @amb26 asked for a unit test for this functionality. Can you provide it before getting Chris to review it?

stegru:
I'm going to leave this one as-is, and move the volume control in-process later: GPII-4445.

PR #273, opened 2019-07-18
GPII-4033: Removed language constraints when dealing with wmi queries

javihernandez:
See https://issues.gpii.net/browse/GPII-4033

### DIFF
--- a/gpii/node_modules/wmiSettingsHandler/src/WmiSettingsHandler.js
+++ b/gpii/node_modules/wmiSettingsHandler/src/WmiSettingsHandler.js
@@ -129,7 +129,9 @@ windows.wmi.updateWMISetting = function (query) {
     try {
         result = updateQuery(query, true);
     } catch (error) {
-        fluid.fail("windows.wmi.updateWMISetting: Failed with WMI error msg - '" + error + "'");
+        if (error.ErrorCode !== "NotSupported") {
+            fluid.fail("windows.wmi.updateWMISetting: Failed with WMI error msg - '" + error + "'");
+        }
     }
 
     return result;
@@ -328,7 +330,11 @@ gpii.windows.wmi.getQuery = function (query) {
     try {
         result = getQuery(query, true);
     } catch (error) {
-        fluid.log("gpii.windows.wmi.getQuery: Query failed with WMI error msg - '" + error + "'");
+        if (error.ErrorCode !== "NotSupported") {
+            // We do not raise an error here because this code is being called by the deviceReporter,
+            // which at this moment doesn't handle failures in the same way the settingsHandler code does.
+            fluid.log("gpii.windows.wmi.getQuery: Query failed with WMI error msg - '" + error + "'");
+        }
     }
 
     return result;
--- a/gpii/node_modules/wmiSettingsHandler/src/WmiSettingsHandler.js
+++ b/gpii/node_modules/wmiSettingsHandler/src/WmiSettingsHandler.js
@@ -129,9 +129,7 @@ windows.wmi.updateWMISetting = function (query) {
     try {
         result = updateQuery(query, true);
     } catch (error) {
-        if (error.message.indexOf("Not supported") === -1 ) {
-            fluid.fail("windows.wmi.updateWMISetting: Failed with WMI error msg - '" + error + "'");
-        }
+        fluid.fail("windows.wmi.updateWMISetting: Failed with WMI error msg - '" + error + "'");
     }
 
     return result;
@@ -330,9 +328,7 @@ gpii.windows.wmi.getQuery = function (query) {
     try {
         result = getQuery(query, true);
     } catch (error) {
-        if (error.message.indexOf("Not supported") === -1 ) {
-            fluid.fail("gpii.windows.wmi.getQuery: Query failed with WMI error msg - '" + error + "'");
-        }
+        fluid.log("gpii.windows.wmi.getQuery: Query failed with WMI error msg - '" + error + "'");
     }
 
     return result;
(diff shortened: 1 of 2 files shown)


javihernandez:
@JavierJF @stegru can you check that this tiny change makes sense?

stegru:
It's beautiful

gpii/node_modules/wmiSettingsHandler/src/WmiSettingsHandler.js
```diff
@@ -330,9 +328,7 @@ gpii.windows.wmi.getQuery = function (query) {
     try {
         result = getQuery(query, true);
     } catch (error) {
-        if (error.message.indexOf("Not supported") === -1 ) {
-            fluid.fail("gpii.windows.wmi.getQuery: Query failed with WMI error msg - '" + error + "'");
-        }
+        fluid.log("gpii.windows.wmi.getQuery: Query failed with WMI error msg - '" + error + "'");
```
amb26: Why did this one get converted to a fluid.log whereas the other one stayed a fluid.fail?

JavierJF: @amb26 you are right, the other one should be changed to fluid.log too.

javihernandez: I converted this one to fluid.log because the deviceReporter is not handling such failures and causes a horrible failure in the system. I left the other one as fluid.fail because it is supposed to be called by the settings handlers, which do handle the exceptions. Is that correct?

amb26: It's correct if you leave comments to that effect : P

javihernandez:
> CI job failed: https://ci.gpii.net/job/windows-tests/1286/

```
Build was aborted
Aborted by unknown
Finished: ABORTED
```

ok to test

javihernandez:
> CI job failed: https://ci.gpii.net/job/windows-tests/1288/

Same issue with the acceptance tests in https://github.com/GPII/windows/pull/311#issuecomment-611469403

ok to test

javihernandez:
> CI job failed: https://ci.gpii.net/job/windows-tests/1290/

```
Vagrant can't use the requested machine because it is locked! This
means that another Vagrant process is currently reading or modifying
the machine. Please wait for that Vagrant process to end and try
again. Details about the machine are shown below:

Name: default
Provider: virtualbox
Build step 'Execute shell' marked build as failure
Finished: FAILURE
```

ok to test

javihernandez:
earth calling CI, you there? I see you're there, doing nothing

image

ok to test

javihernandez:
ok to test

javihernandez:
> CI job failed: https://ci.gpii.net/job/windows-tests/1295/

Weird failure

image

It claims that the acceptance tests timed out after 90 mins but in the logfile we can see that that there's only 4 minutes between the first and the last logline, see:

```
06:45:26.208:  Registering module gpii-windows from path \\vboxsvr\vagrant\
...
06:49:26.130:  UserLogonRequest completes successfully with the payload: User with GPII key os_win was successfully logged in.
```

ok to test

javihernandez:
> CI job failed: https://ci.gpii.net/job/windows-tests/1300/

Unrelated acceptance test failure,

```
11:24:22.912:  jq: Expected: {
    "gpii.windows.enableRegisteredAT": {
        "com.microsoft.windows.magnifier": [
            {
                "settings": {
                    "running": true
                }
            }
        ],
        "com.microsoft.windows.onscreenKeyboard": [
            {
                "settings": {
                    "running": false
                }
            }
        ]
    }
}
11:24:22.912:  jq: Actual: {
    "gpii.windows.enableRegisteredAT": {
        "com.microsoft.windows.magnifier": [
            {
                "settings": {
                    "running": undefined
                }
            }
        ],
        "com.microsoft.windows.onscreenKeyboard": [
            {
                "settings": {
                    "running": false
                }
            }
        ]
    }
}
```
ok to test

javihernandez:
@stegru all yours

gpii/node_modules/wmiSettingsHandler/src/WmiSettingsHandler.js
```diff
@@ -129,7 +129,7 @@ windows.wmi.updateWMISetting = function (query) {
     try {
         result = updateQuery(query, true);
     } catch (error) {
-        if (error.message.indexOf("Not supported") === -1 ) {
+        if (error.ErrorCode !== "NotSupported") {
             fluid.fail("windows.wmi.updateWMISetting: Failed with WMI error msg - '" + error + "'");
```
stegru, line 133: You might want to replace `error` with `error.message` (or put it in a separate argument?), otherwise we'll get `[object Object]` in the log. Same goes for the one on line 336.

Also, line 336 has been changed to `fluid.log`, but this is still `fluid.fail` - just checking.

javihernandez: When having the failure, it produces the following output:

`02:48:28.931:  jq: FAIL: Module "WmiSettingsHandler Module" Test name "Testing wmiSettingsHandler set handles null/non null oldValue" - Message: Assertion failure (see console.log for expanded message): windows.wmi.updateWMISetting: Failed with WMI error msg - 'System.Management.ManagementException: Not supported '
`
In this case, we use fluid.fail because this code is called by the settingsHandler, which will "gracefully" handle such failures.

Regarding line 336, we use fluid.log because that function is called from the deviceReporter (see https://github.com/GPII/universal/blob/ad461d971122af7df00ae4411cdf353f3d90bd70/testData/solutions/win32.json5#L4806-L4812), which doesn't "gracefully" handle such failures and makes the system to crash. By using fluid.log, we just inform about that and the solution is not going to be reported as installed.

stegru: ah yes, it's a .NET exception that has a `toString` implementation.

PR #276, opened 2019-08-16
GPII-4087: Added hack to stop wallpaper being reset.

stegru:
When resetting to standard, the high-contrast is being turned off, even though it's already turned off.

This makes Windows reload the last save theme, causing the custom wallpaper to be lost.

This change just checks if the high-contrast is already on before setting it on. It's assumed this type of check is performed before invoking the settings handler - but for an unknown reason it's not. This is a quick low-risk isolated fix for the high-contrast.

To test: Change the wallpaper, start gpii, key in as `reset`.


javihernandez:
Hey @stegru, nice idea - However, I'm getting an error right after having reset to std. Shouldn't this be more correct? -> https://pastebin.com/jiY22mJe

stegru:
No, because if high-contrast is on that wouldn't turn it off.

What's the error?

stegru:
ah

stegru:
At the moment, this PR contains a hack that turns `windows.rm` into a no-op, so a cloud update is not required.

see also: https://github.com/stegru/windows/pull/11

amb26:
Notes from meeting on 21/8/19 - this pull request must never be merged into master since its strategy is to corrupt the API of gpii.windows.rm to ensure that it is neutered with respect to the current solutions registry.
Once our staging pipeline is clear, the next step is to update the solutions registry so that the rm blocks are removed - this will not break existing clients - and then we can issue an updated version of this pull which just culls out the calls to rm and roll it out to clients.

amb26:
07:22:34.994:  ASSERTION FAILED: Cannot look up name gpii.windows.rm to a function with registered argumentMap - got defaults undefined
07:22:34.994:  Current activity: 
07:22:34.994:      while invoking invoker with name executeActions and record {
    "func": undefined,

javihernandez:
> 07:22:34.994: ASSERTION FAILED: Cannot look up name gpii.windows.rm to a function with registered argumentMap - got defaults undefined
> 07:22:34.994: Current activity:
> 07:22:34.994: while invoking invoker with name executeActions and record {
> "func": undefined,

I'd say that this PR needs to pull a version of universal where gpii.windows.rm is removed

stegru:
Yes, forgot to push the universal change.

ok to test

amb26:
Hi there @stegru - I've rolled a new universal build at 0.3.0-dev.20200511T123345Z.a793adf
but consistently get an error on the windows tests with 
````
07:47:48.340:  jq: FAIL: Module "gpii.tests.metrics" Test name "Testing application metrics" - Message: Error: ENOENT: no such file or directory, open 'C:\Users\vagrant\AppData\Local\Temp\gpii
-test-metrics-1589208467904'
07:47:48.340:  jq: Source:     at pok (V:\node_modules\infusion\tests\test-core\jqUnit\js\jqUnit.js:112:15)
    at Object.fail (V:\node_modules\infusion\tests\test-core\jqUnit\js\jqUnit.js:129:13)
    at ReadStream.<anonymous> (V:\gpii\node_modules\windowsMetrics\test\WindowsMetricsTests.js:860:16)
    at ReadStream.emit (events.js:182:13)
    at lazyFs.open (internal/fs/streams.js:115:12)
    at FSReqWrap.oncomplete (fs.js:141:20)
````
can you help?

amb26:
In theory all that is in the universal build is the removal of gpii.windows.rm

PR #277, opened 2019-08-20
GPII-4092: QSS failing due to no sound card.

stegru:
Quick fix to stop the QSS from failing to load, due to a rejection in the volume handler when there's no sound card.


javihernandez:
ok to test

javihernandez:
ok to test

stegru:
ok to test

PR #284, opened 2019-11-18
GPII-4231: Onboard color vision filters

JavierJF:
Branch based on GPII-4152 for testing new onboarded color filters.

**Warning do not merge**


provisioning/NpmInstall.ps1
```diff
@@ -12,10 +12,14 @@ Import-Module (Join-Path $scriptDir 'Provisioning.psm1') -Force -Verbose
...
 $msbuild = Get-MSBuild "4.0"
 
-# Build the settings helper
+# Build the settingsHelper solution
+nuget restore .\settingsHelper\SettingsHelper.sln
+
+$env:VisualStudioVersion = '15.0'
 $settingsHelperDir = Join-Path $rootDir "settingsHelper"
-Invoke-Command $msbuild "SettingsHelper.sln /p:Configuration=Release /p:Platform=`"Any CPU`" /p:FrameworkPathOverride=`"C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETFramework\v4.5.1`"" $settingsHelperDir
+Invoke-Command $msbuild "SettingsHelper.sln /p:Configuration=Release /p:Platform=`"x64`"" $settingsHelperDir
```
stegru: There's a problem with this - `x64` will prevent it from working on a 32-bit OS. (We can't make it `x86`, because then it wont work on a 64-bit OS).

Two choices:
- Do nothing if running 32-bit Windows
- Create a 32-bit version, and execute that for 32-bit Windows

We don't test 32-bit in CI.. so the safest + quickest option would be to do nothing for 32-bit (confirm with @amb26) - also, according to the metrics there are no 32-bit installations.

(OS bit detection: `is64 = os.arch() === "x64" || windows.isWow64()`)

stegru: If compiling for `x86` does work on 64-bit, then just keep it to `x86` - but that would surprise me.

settingsHelper/README.md
```diff
@@ -1,99 +0,0 @@
-# Windows system settings handler helper process
```
stegru: Should update this file, rather than delete it.

settingsHelper/SettingsHelperApp/SettingsHelperApp.cpp
```diff
@@ -0,0 +1,32 @@
...
+using std::pair;
+
+#pragma comment (lib, "WindowsApp.lib")
+
+[System::STAThread]
+int wmain(int argc, wchar_t* argv[]) {
+    pair<int, wchar_t**> payload { argc, argv };
+    DWORD threadID { 0 };
+    HANDLE thHandle { NULL };
+
```
stegru: Add a comment to explain why this needs to be in a separate thread - I'm curious

settingsHelper/SettingsHelperLib/BaseSettingItem.cpp
```diff
@@ -0,0 +1,254 @@
+#include "stdafx.h"
```
stegru: Don't forget the copyright header, explaining what this file does.

settingsHelper/SettingsHelperLib/BaseSettingItem.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+            if (res == ERROR_SUCCESS) {
+                UINT it = 0;
+                UINT wait = timeout / retries;
+
+                while (isUpdating == TRUE && res == ERROR_SUCCESS) {
+                    System::Threading::Thread::Sleep(10);
+                    res = this->setting->get_IsUpdating(&isUpdating);
+
+                    if (res != ERROR_SUCCESS) {
```
stegru: This check is already in the `while` expression

settingsHelper/SettingsHelperTests/SettingItemTests.cpp
```diff
@@ -0,0 +1,578 @@
...
+        ATL::CComPtr<IInspectable> value;
+        res = colSetting.GetValue(settingIds[2], value);
+        EXPECT_EQ(res, ERROR_SUCCESS);
+        EXPECT_TRUE(value != NULL);
+    }
+}
+
+/*
+/// <summary>
+///  NOTE: This test should remain commented, as it's only used for development purposes.
```
stegru: `#if 0` may be better?

settingsHelper/SettingsHelperTests/SettingUtilsTests.cpp
```diff
@@ -0,0 +1,185 @@
...
+using std::pair;
+using std::wstring;
+
+#pragma comment (lib, "WindowsApp.lib")
+#include <metahost.h>
+#pragma comment(lib, "mscoree.lib")
+
+using namespace ABI::Windows::Foundation;
+
+TEST(CreatePropertyValue, CreateDateFromDouble) {
```
stegru: There are quite a lot of un-commented tests

stegru:
Is this a drop-in replacement for the original c# implementation? There are no js code changes?

provisioning/Chocolatey.ps1
```diff
@@ -30,4 +30,20 @@ Invoke-Command $chocolatey "install python2 -y"
...
+refreshenv
+
+# Install the required ATL Library & WindowsSDK for SystemSettingsHandler
+Invoke-Command $chocolatey 'install --force -y vcbuildtools -ia "/InstallSelectableItems VisualCppBuildTools_ATLMFC_SDK;VisualCppBuildTools_NETFX_SDK"'
+refreshenv
+
+Invoke-Command $chocolatey "install windows-sdk-10-version-1809-all --version=10.0.17763.1 -y"
+refreshenv
+
+Invoke-Command $chocolatey "install windows-sdk-8.1 -y"
```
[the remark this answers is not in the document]
JavierJF: yes it does, that's why I make the re-installation here.

provisioning/Chocolatey.ps1
```diff
@@ -30,4 +30,20 @@ Invoke-Command $chocolatey "install python2 -y"
 Add-Path $python2Path $true
 refreshenv
 
+Invoke-Command $chocolatey "install nuget.commandline"
+refreshenv
+
+Invoke-Command $chocolatey "uninstall windows-sdk-8.1 -y"
```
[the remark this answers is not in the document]
JavierJF: Yes, this is commented in the other pull, GPII-4152, the only reason this is done like this, it's to be sure through CI that it's safe to do this in the base VM images.

provisioning/NpmInstall.ps1
JavierJF: @stegru it doesn't work at all if you compile it for `x86,` I would need to verify if everything remains the same using a `x86` virtual machine.

JavierJF:
@stegru yes, it aims to be a replacement of the original C# implementation, no js code changes are required, the interface of the old handler is preserved.

settingsHelper/SettingsHelperTests/SettingUtilsTests.cpp
JavierJF: @stegru yes, most of the test were very unitary so I didn't add to much comments there, will be fixing it.

settingsHelper/SettingsHelperLib/BaseSettingItem.cpp
JavierJF: Yes, and I think that is better to use the one on the while because in case of failure releases more stuff **explicitly**.

settingsHelper/SettingsHelperTests/SettingItemTests.cpp
JavierJF: Yes, I like that more, will change it.

settingsHelper/SettingsHelperApp/SettingsHelperApp.cpp
JavierJF: I did this in order to be able to have a global waiting for the main thread if necessary, because as you know, there have been some issues in the past with dll unloading when the dependencies between them were unknown, this way I can shutdown the whole application if the thread doesn't shutdown itself after an specified timeout. That **right now** is not necessary, but who knows if we could need it in the future.

stegru:
ok to test

provisioning/Chocolatey.ps1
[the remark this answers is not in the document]
amb26: Horrours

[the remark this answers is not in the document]
stegru: Is it just code in the PR that fails without 8.1? If so, I suggest fixing that. I think using a single SDK would be a better idea.

(I've built master using v10 without a problem)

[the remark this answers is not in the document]
amb26: @JavierJF - could you check this out?

settingsHelper/SettingsHelperLib/IPropertyValueUtils.cpp
```diff
@@ -0,0 +1,339 @@
+/**
+ * Utilities for handling IPropertyValue types.
```
amb26: Can you explain in more details what the strategy of this file is? Where did all the code come from?

settingsHelper/SettingsHelperTests/ParsingTests.cpp
```diff
@@ -0,0 +1,112 @@
+/**
+ * Tests for the parsing functions.
```
amb26: What, if anything, executes these tests?

provisioning/Chocolatey.ps1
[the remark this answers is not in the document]
JavierJF: @amb26 I actually did! In the current version of the script it's fixed, the problem was the order of installation of the SDK and node development packages, the 8.1 SDK is in fact, not needed.

settingsHelper/README.md
JavierJF: Missed this one, I will update it with a brief description of what the handler does.

settingsHelper/SettingsHelperLib/IPropertyValueUtils.cpp
JavierJF: The purpose of the fails is to provide some helper functions for dealing with IPropertyValue types,  since only the interface of the type is accessible from the native API, doing simple operations with them like construction or comparing them is tedious. The purpose of this file is to hold those helper functions.

settingsHelper/SettingsHelperTests/ParsingTests.cpp
JavierJF: @amb26 nothing is executing this tests right now, I created them during development for being sure of not breaking anything, the handler is exercised with end-to-end tests, but I agree that we could do something for adding this unitary tests to the global test execution. Do you think I should add them? And in that case, should I place it's launching in the ```testSystemSettingsHandler.js``` as for the end-to-end tests?

amatas:
ok to test

JavierJF:
image

A timeout failure. Testing again.

JavierJF:
Ok to test.

amatas:
ok to test

amatas:
ok to test

amatas:
ok to test

The provision of the VM takes more than 45 minutes and it reaches the timeout.

amb26:
ok to test

amb26:
Timeout from NuGet on installing node.js
````
    default: Chocolatey v0.10.15
    default: Installing the following packages:
    default: nodejs.install
    default: By installing you accept licenses for the packages.
    default: [NuGet] The operation has timed out
````

amb26:
ok to test

amb26:
ok to test

settingsHelper/SettingsHelperTests/ParsingTests.cpp
amb26: Yes, please cause the tests to actually run!

settingsHelper/SettingsHelperLib/IPropertyValueUtils.cpp
amb26: Please reflect this info the comments

settingsHelper/SettingsHelperApp/SettingsHelperApp.cpp
amb26: Please add the comment!

settingsHelper/SettingsHelperLib/IPropertyValueUtils.cpp
JavierJF: Done! Now there is a comment in the header reflecting the purpose of the file.

settingsHelper/SettingsHelperTests/ParsingTests.cpp
JavierJF: Done! Tests are executed and parsed in search for fails.

javihernandez:
Replaced with https://github.com/GPII/windows/pull/311

PR #155, opened 2017-10-20
GPII-2338: Windows service.

stegru:
Here's the Windows Service, that starts GPII and the listeners when a user logs into windows, and restarts them if they die.

At the moment, this provides no value; GPII will operate the same as before. ~~The installer needs to be adjusted to install and start the service.~~

More info in the [README](https://github.com/stegru/windows/blob/GPII-2338/service/README.md).

**Update:**
This changes the way GPII is started. Instead of using the AutoRun registry key, the service starts GPII when the user logs on. Its passed an environment value `GPII_SERVICE_PIPE` which instructs gpii to connect to the pipe (GPII-2399).

If it's ok if GPII is started normally - it just wont be able to enjoy the goodness of the service (currently nothing but restart on crash). However, if the service is started from within the source code (ie, via node service/index.js), then it will open an insecure pipe to which GPII will connect.

The service writes a log to `C:\ProgramData\GPII\gpii-service.log` (`%APPDATA%/GPII/` is not used, because that's the user's profile directory, and the service doesn't run as the user)

PR #287, opened 2019-12-03
GPII-3572/GPII-4244: Metrics switch, and site-config updates

stegru:
Makes the service automatically updates the site-config file (GPII-4244). It also reads the `metricsSwitch` value from the siteConfig and starts Morphic with the appropriate `NODE_ENV` environment variable.

Configurable here (disabled by default): https://github.com/GPII/gpii-app/blob/2e99c3f5db3bcd236737b623e287361049d9029c/provisioning/service.json5#L18

Works with https://github.com/GPII/gpii-app/pull/73

This includes the GPII-2338 branch (https://github.com/GPII/windows/pull/155), the changes related to this PR are https://github.com/stegru/windows/compare/GPII-2338...stegru:GPII-4244?expand=1


gpii-service/src/processHandling.js
```diff
@@ -113,11 +115,54 @@ processHandling.getProcessList = function (allProcesses) {
...
+
+        service.logDebug("Setting desktop icon: ", desktop, stashed, morphicHide ? "hide" : "show");
+
+        try {
+            // Ensure there's a copy in the stash location
+            if (!fs.existsSync(stashed)) {
+                fs.copyFileSync(desktop, stashed);
+            }
+
+            if (morphicHide) {
```
javihernandez: I guess we also need a "morphicShow" - this is, to ensure that the shortcuts are copied back to the desktop when we switch from off to on.

stegru: oops

javihernandez:
ok to test

gpii-service/package.json
```diff
@@ -19,11 +19,14 @@
     "dependencies": {
         "ffi-napi": "2.4.4",
         "json5": "2.1.0",
+        "mkdirp": "0.5.1",
         "minimist": "1.2.0",
-        "@gpii/os-service": "2.2.0",
+        "@gpii/os-service": "stegru/node-os-service#GPII-3934",
```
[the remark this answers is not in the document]
stegru: That change wasn't intended for this branch - reverted it.

gpii-service/src/service.js
```diff
@@ -59,6 +59,31 @@ service.logImportant = logging.important;
...
 
+/**
+ * @typedef {Object} Config
+ * @property {Object<String,ProcessConfig>} processes Child process.
+ * @property {Object} logging Logging settings
+ * @property {String} secretFile The file containing site-specific information.
+ * @property {String} package.json The package.json file for gpii-app.
+ * @property {String} siteConfigFile The site-config file.
+ * @property {AutoUpdateConfig} autoUpdate Auto update settings.
+ * @property {String} morphicVersion [generated] The version field from gpii-app's package.json.
```
amb26, line 70: @javihernandez - looks like we are not actually publishing these to npm via fluid-publish or whatever as part of our release process - https://www.npmjs.com/package/gpii-app - I assume we need to actually do this before this facility becomes useful?

gpii-service/README.md
```diff
@@ -102,7 +102,68 @@ This gets installed in `c:\Program Files (x86)\Morphic\windows\service.json5`.
...
-    "secretFile": "%ProgramData%\\Morphic Credentials\\secret.txt"
+    "secretFile": "%ProgramData%\\Morphic Credentials\\secret.txt",
+    // The gpii-app package.json file (default: "resources/app/package.json")
+    "package.json": "resources/app/package.json",
+
+    // Auto update of files
+    "autoUpdate": {
+        "enabled": false, // true to enable
+        // Where to store the 'last update' info
+        "lastUpdatesFile": "%ProgramData%\\Morphic\\last-updates.json5",
```
amb26, line 113: Shouldn't this more reasonably go in the morphic settings directory? Is there an expander syntax that will allow that?

gpii-service/src/service.js
javihernandez: Actually, the machinery involved is:

1. Whatever is in the version field of the gpii-app's package.json file. No matter whether is published in npm or not.
2. https://github.com/GPII/gpii-morphic-site-resources, and when a new release is out, we must update the defaults (and any other site-specific defaults) according to the version that we have just released.

This is the approach we took when designing this, but given that there are no more pilot sites, the autoUpdate policy was turned on by default (see https://github.com/GPII/gpii-app/blob/master/provisioning/service.json5#L20). If by any chance we want to make this functionality work, we'd need to update the site-resources repo as indicated above and ship a service.json5 file with the autoUpdate functionality enabled.

amb26: OK - but I'm assuming that this pull is not part of your "secret sauce" 1.3.x branches so far, right? Despite it being marked as "@@@@" in the pad, when you say, "If by any chance we want to make this work" it sounds like it is not actually in the high priority list for delivered functionality at the very least for the case studies - does it sound like we want to turn this off by default before we merge this pull, just to make sure it doesn't destabilise anything and we can opt into it later if necessary?

javihernandez: IT IS part of 1.3.x, but we're not using the _autoUpdate_ functionality right now, it's disabled [by default](https://github.com/GPII/gpii-app/blob/master/provisioning/service.json5#L20), so we're fine to go. Note that we've used the 4 state metrics switch functionality from this PR, which "I think" is what was considered as a priority ("@@@@" in the pad).

PR #290, opened 2019-12-13
GPII-3810: Reverting to current released version of universal

amb26:


amb26:
Stalled after 5 minutes, no clear reason
ok to test

PR #295, opened 2020-02-16
GPII-4355: Updated build scripts and projects to use VS2017 build tools

christopher-rtf:
This PR, combined with the gpii-app/GPII-4355 PR, should fully enable Morphic for Windows to compile using modern (VS2017) developer tools and Node.js v10.16.3.

Please note that this branch is based on the current "master" branch of GPII/windows (which contains a bug which is preventing most of the "set" Settings Handlers from working).  However it should be able to be merged with any recent production-quality commit in the master branch.

Critical: developers must upgrade their developer environment to Visual Studio 2017.  The new build scripts use Microsoft's new "vswhere" utility (included with VS2017+) to locate MSBuild, CSC, environment variables, etc.  This is a hard fork in the road...and we will need all the developers on our team (plus the CI environment) to move to VS2017 when this goes "live" in master.

For the future: The $visualStudioVersion PowerShell variable has been designed to support either a specific version (e.g. "15.0" for VS2017) or a range (e.g. "[15.0,16.0)" as an inclusive range for any VS2017'ish releases from v15.0 to v15.9999999999).  This means that as we want to start adding support for VS2019 or other newer releases, we can simplly update the $visualStudioVersion variable to include a wider range of compiler versions.  Our build may indeed be compatible with VS2019 as well, but I have not tested with VS2019 due to the fact that Electron still uses VS2017 in their documentation.

### DIFF
--- a/gpii/node_modules/nativeSettingsHandler/nativeSolutions/VolumeControl/VolumeControl/VolumeControl.vcxproj
+++ b/gpii/node_modules/nativeSettingsHandler/nativeSolutions/VolumeControl/VolumeControl/VolumeControl.vcxproj
@@ -28,26 +28,26 @@
   <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'" Label="Configuration">
     <ConfigurationType>Application</ConfigurationType>
     <UseDebugLibraries>true</UseDebugLibraries>
-    <PlatformToolset>v140</PlatformToolset>
+    <PlatformToolset>v141</PlatformToolset>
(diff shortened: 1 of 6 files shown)


amb26:
ok to test

christopher-rtf:
While we're waiting to update CI to install VS2017 build tools, here is the log of a manual checkout/build from a local VM.  [Build successful]

[windows_build_log_vs2017.txt](https://github.com/GPII/windows/files/4210095/windows_build_log_vs2017.txt)

amb26:
ok to test

amb26:
ok to test

PR #297, opened 2020-02-21
GPII-4365: Language test improvements

stegru:
Improved the error reporting in the language tests.

Also stopped gpii from being started by fixing the main script in `package.json` and removing the `%gpii-windows` requires from the tests (they're harmless, but not needed).


stegru:
ok to test

PR #298, opened 2020-02-25
NOJIRA: Building the service in its own isolated drive.

stegru:
This will stop the `gpii-service` build from taking things from `/node_modules`, to using its own `node_modules` instead.


stegru:
ok to test

amb26:
ok to test

PR #301, opened 2020-02-28
GPII-4370: Rejecting settings handler if the settings helper fails.

stegru:
This stops Morphic from stalling and preventing the logout/shutdown of Windows.


amb26:
Looks highly sensible and something that should have been caught during code review : P

PR #306, opened 2020-03-23
GPII-4125: Update to work with initial implementation of the LSR.

the-t-in-rtf:
@amb26, this is a companion pull to test [the work in progress on the LSR](https://github.com/GPII/universal/pull/860).  It should only be merged if changes (path to SR files, etc.) are required.


the-t-in-rtf:
There was a failure in the postinstall step for gpii-universal that I believe I've fixed.  On a local machine I get one test failure related to the resolution.  OK to test again, let's see if that's just a byproduct of my local environment or a real failure.

the-t-in-rtf:
Ok to test again following recent updates to the related universal pull.

the-t-in-rtf:
The latest issues in local testing are with the screen resolution, which is reported as 1024x768 when it should be 800x600.  I'm a bit baffled as to why, my generated schema for that matches what's passing in [311l](https://github.com/GPII/windows/pull/311).  Ok to test again here to see if it's unique to my local environment.

the-t-in-rtf:
The last build was aborted, I never see anything like this locally.  Ok to test again.

the-t-in-rtf:
With the screen mirror setting removed, this should now pass.  Ok to test again.

the-t-in-rtf:
There is no need for this to exist as a pull, there may not even be working CI shortly.

PR #309, opened 2020-03-29
GPII-4420: Only listing languages installed by a morphic language msi

stegru:
The list of available languages is sometimes incorrect.

This change will make the UI only list those languages installed by Morphic language installers.


javihernandez:
ok to test

javihernandez:
ok to test

PR #313, opened 2020-04-16
GPII-4446: Modifying the existing ribbon, rather than replacing it.

stegru:
Instead of swapping the UI XML files, this will extract the relevant nodes from a given document and put them into the existing UI.

If a tab already exists, then only the `visibility` attribute is applied.

This should prevent users from losing their customisations (either made before or after applying the setting).


amb26:
Do update this to master and try it out, I did this in https://github.com/GPII/windows/pull/317/files and ran into an odd test failure with the metrics

PR #322, opened 2020-05-28
GPII-4470: Storing the currently active window in an environment variable

stegru:
This just updates an environment variable, so a new child process knows the active window before the QSS was clicked.

This is so the select to speech app (which will be called by a BYOB) knows what window to read the selected text from.


stegru:
ok to test

```
03:46:54.403:  jq: FAIL: Module "gpii.tests.acceptance.windows.builtIn.config tests" Test name "Testing os_win using default matchmaker" - Message: Error making request to /user/os_win/login: socket hang up
```

PR #327, opened 2020-06-16
Node12+ref napi

javihernandez:
This is just a test to see if the replacement of ref with ref-napi produces the same problems that I'm getting locally on CI


gpii-service/package.json
```diff
@@ -21,7 +21,7 @@
         "json5": "2.1.0",
         "mkdirp": "0.5.1",
         "minimist": "1.2.0",
-        "@gpii/os-service": "2.2.0",
+        "@gpii/os-service": "javihernandez/node-os-service#GPII-4140",
```
amb26, line 24: What's in this one?

javihernandez: It's an update with upstream/master, which adds support for node12, see: https://github.com/GPII/node-os-service/pull/2

PR #329, opened 2020-11-16
GPII-4517: Added workflow for github actions

javihernandez:
Here's an example of one CI pass https://github.com/javihernandez/windows/runs/1406430176?check_suite_focus=true

### DIFF
--- a/.github/workflows/node-tests.yml
+++ b/.github/workflows/node-tests.yml
@@ -31,4 +31,9 @@ jobs:
       - run: npm config set python python2.7
 
       - run: npm install
-      - run: node tests/UnitTests.js
+
+      - name: Pass linter
+        run: $(npm bin)/grunt lint
+
+      - name: Run node unit tests
+        run: node tests/UnitTests.js
--- a/.github/workflows/node-tests.yml
+++ b/.github/workflows/node-tests.yml
@@ -15,6 +15,7 @@ jobs:
           node-mirror: "https://nodejs.org/download/release/"
 
       - uses: actions/checkout@v2
+      - run: ./provisioning/CouchDB.ps1
       - uses: actions/setup-python@v2
         with:
           python-version: '2.7.17'
@@ -37,3 +38,8 @@ jobs:
 
       - name: Run node unit tests
         run: node tests/UnitTests.js
+
+      - name: Run Acceptance builtIn tests
+        env:
+          GPII_TEST_COUCH_USE_EXTERNAL: 1
+        run: node tests/AcceptanceTests.js builtIn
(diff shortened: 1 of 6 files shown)


.github/workflows/node-tests.yml
```diff
@@ -0,0 +1,39 @@
...
+      - run: npm config set msvs_version 2015 --global
+      - run: npm config set python python2.7
+
+      - run: npm install
+
+      - name: Pass linter
+        run: npx grunt lint
+
+      - name: Run node unit tests
+        run: node tests/UnitTests.js
```
amb26, line 40: Worth trying a few acceptance tests? Do they run?

javihernandez: They would run, but if we want to run them as they are we'd need to set up couchdb. I can definitely take a look if there's an easy way to set up couchdb in the github-hosted windows images. I think I can even take a look at creating a small github action that sets up couchdb on windows (I already did [something similar](https://github.com/GPII/windows/blob/master/provisioning/CouchDB.ps1) for our vagrant boxes), but again, I'd need some time.

Another option would be to run the acceptanceTests against a file-based prefsServer, which I think it's not the way we want to test things.

javihernandez: So yeah, let me give a try at setting up couchdb, I'll keep you posted :wink:

javihernandez: Done, here's the link to the CI pass https://github.com/javihernandez/windows/runs/1415137751?check_suite_focus=true

package.json
```diff
@@ -19,7 +19,7 @@
     "dependencies": {
         "edge-js": "10.3.1",
         "ffi-napi": "2.4.3",
-        "gpii-universal": "0.3.0-dev.20200604T131823Z.d6fed9a",
+        "gpii-universal": "javihernandez/universal#f2972b9deeae48847a4ed951d1ab703b4e7bafc5",
```
amb26: Which commit is this - should we roll a release of it?

javihernandez: Right, the universal bits are in this PR https://github.com/GPII/universal/pull/886
So let's say this PR needs that one the get merged first

javihernandez:
CI run pass here: https://github.com/javihernandez/windows/runs/1441876960?check_suite_focus=true

amb26:
What's the state of the CI at this end? I see a "Pending - Build triggered" note from the bot, but I assume this form of CI is gone. What's required to get the github actions CI as the official one for this repo?

javihernandez:
Yes, this is the old CI, which is gone now. I already checked that further PRs won't trigger any job into the old CI, see https://github.com/GPII/windows/pull/330

After merging this PR, the tests will run on github actions with every new PR and whenever a new commit lands into master. All CI passes will appear in https://github.com/GPII/windows/actions as soon as this is merged as it happened in universal, see https://github.com/GPII/universal/actions

PR #330, opened 2020-11-23
test that the old ci is gone

javihernandez:


javihernandez:
The old CI is not triggered anymore

